PR #711, opened 2015-01-24
sync migration admin password with config/ion_auth.php

alpinsky:
The default login/password specified in README.md currently is not working after applying 001_install_ion_auth.php. The admin password need to be changed in migration since default hash_method is bcrypt. Default configuration should work.


benedmunds:
Closing due to being stale.

PR #718, opened 2015-01-28
Group permissions

sadika9:
Add new role based (like on Spring) feature.

Now each group has one or many permissions assigned. When member is assigned with a group all permissions of that group will be assigned to that user. And we can check whether user has sufficient permission to perform some task by using has_permission() method.

e.g. has_permission('create_new_user')

Also It would be enough to assign a user only to a single group because several groups can be created which have different access permissions. But I did not change database schema (still has many-to-many) & logic because I need to keep this pull request simple.

I don't have access progress & ms sql databases so I did not update those sql files.

DB Design:
ion_db_new

### DIFF
--- a/config/ion_auth.php
+++ b/config/ion_auth.php
@@ -31,6 +31,8 @@ $config['tables']['users']           = 'users';
 $config['tables']['groups']          = 'groups';
 $config['tables']['users_groups']    = 'users_groups';
 $config['tables']['login_attempts']  = 'login_attempts';
+$config['tables']['permissions']        = 'permissions';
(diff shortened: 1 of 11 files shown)


avenirer:
interesting idea...

benedmunds:
Hey,

I appreciate this, looks like a ton of work and very well done.  I feel that ACL expands the scope of Ion Auth past what is needed for the majority of users though.  

Maybe look at using hooks to extend Ion Auth and then offering this as your own package to add ACL on top of Ion Auth?

PR #721, opened 2015-01-30
destroy and regenerate in PHP

avenirer:
destroy session and regenerate session directly in PHP

### DIFF
--- a/libraries/Ion_auth.php
+++ b/libraries/Ion_auth.php
@@ -393,7 +393,8 @@ class Ion_auth
 		}
 
 		//Destroy the session
-		$this->session->sess_destroy();
+		session_destroy();
+		
 
 		//Recreate the session
 		if (substr(CI_VERSION, 0, 1) == '2')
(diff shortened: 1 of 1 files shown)


benedmunds:
Whats the reason for bypassing CI?

benedmunds:
and does this work with something like CI's DB session handling.

avenirer:
At least regarding CI 3, sess_regenerate() is actually an alias for the PHP native function session_regenerate_id (as I see in the manual). Regarding the sess_destroy() method, looking inside the version 2.x's session library I see that is actually calling session_destroy().

I actually do have a question regarding the recreation of the session: Why do you need to recreate it manually? The session is created again automatically when the Session library is loaded. Or am I wrong? I would remove that part entirely.

benedmunds:
If I remember correct regenerating it was so it would go ahead and clear everything, otherwise it was taking one additional load before it cleared and there could be leftover data.  It's been awhile since I wrong that though so I could be wrong.

PR #722, opened 2015-02-01
unnecessary if / else

sv3tli0:
There is no point using: if ( TRUE ) return TRUE else return FALSE.. 
Just return the check it will be bool.

### DIFF
--- a/controllers/auth.php
+++ b/controllers/auth.php
@@ -780,15 +780,8 @@ class Auth extends CI_Controller {
 
 	function _valid_csrf_nonce()
 	{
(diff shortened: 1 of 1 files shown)


benedmunds:
I prefer the old way.  It matches CI's style more and is more explicit.

sv3tli0:
Ok, just I don't see the point to catch bool and after that to return it, instead just to return it.

benedmunds:
Functional it's no different. But conceptually it can be easier to understand, especially for a more junior developer.

PR #724, opened 2015-02-03
Fix typo

Tjoosten:


benedmunds:
<?endif?> should be valid syntax.  Are you getting a syntax error?

Tjoosten:
not tested but i get used to use `endif;`

benedmunds:
Ah, it should be fine then, it's valid short syntax.  Thanks though!

PR #726, opened 2015-02-06
create_user() and update_user() alternate methods

avenirer:
We should keep consistent method names. If you have delete_user(), create_group(), update_group(), delete_group(), we should also have create_user() and update_user().


benedmunds:
Good move, thanks!

Issue #604, opened 2014-07-11
How to change language?


tikky:
Hello,

I have codeigniter set to polish language.
Where in ION AUTH I can change the default language for polish?

tikky:
Long time ago I've asked for a help with polish language.
Today I found possible problem.

If you will take a look for example at italian tranlsation we have two files:
- auth_lang.php
- ion_auth_lang.php
and both are tranlsated to italian.

In Polish language folder we have also two files, but
auth_lang.php - it is on original - english version
ion_auth_lang.php - it is a polish version

So my question is: auth_lang.php - shouldn't be also translated to Polish?
If yes I can do it.

benedmunds:
Yes it does need to be translated.

PR #729, opened 2015-02-09
Translated whole file to Polish language

tikky:
Original file in Polish folder wasn't translated.
So in order to resolve my own issue #604 I have just finished the tranlsation.
https://github.com/benedmunds/CodeIgniter-Ion-Auth/issues/604


benedmunds:
Thank you!

PR #732, opened 2015-02-10
index_user_edit_link

tikky:
Added index_user_edit_link instead of hardcoded "edit"


benedmunds:
See comments in issue #734

PR #733, opened 2015-02-10
Added translation for index_user_edit_link

tikky:


benedmunds:
See comments in issue #734

PR #734, opened 2015-02-10
Added index_user_edit_link to original language file

tikky:


benedmunds:
This will need to be address to all lang files.  And the view should be updated to match.  All in the same PR.

benedmunds:
Closing due to being stale.

Issue #710, opened 2015-01-20
Ability to use group name in groups()


sykosoft:
It's an extra step in our model to get the id of the group and then the members of a group as we are unable to use a group name in groups(), and are required to use a group's ID. If, like in_group(), we could use a group's name, this would greatly simplify certain operations.

benedmunds:
Send a PR and I'll probably merge it.

PR #740, opened 2015-02-19
allow $this->ion_auth->users() to accept group name

avenirer:
As answer to issue #710, allow $this->ion_auth->users() to accept group name or an array of group names as parameter:
For example, now it can also accept `$this->ion_auth->users('members');` or `$this->ion_auth->users(array('members','admin');`

### DIFF
--- a/models/ion_auth_model.php
+++ b/models/ion_auth_model.php
@@ -1308,21 +1308,25 @@ class Ion_auth_model extends CI_Model
 				    'inner'
 				);
 			}
-			// verify if group name or group id was used
-            		$numeric_group = TRUE;
-            		foreach($groups as $group)
-            		{
-                		if(!is_numeric($group)) $numeric_group = FALSE;
-            		}
-            		//if group name was used we do one more join with groups
-            		if($numeric_group==FALSE)
-            		{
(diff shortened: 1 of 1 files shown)


benedmunds:
Looks like a mixed array would break this.  Should probably support both.  Like:

    array('members', 2, 'admin')

avenirer:
You are right... How about now?

benedmunds:
Much better.  What's going on here:

$this->db->{$or_where_in}

avenirer:
Well... I thought it would be better to generate that method dinamically. Only if both arrays are not empty that variable will be transformed into $this->db->or_where_in(). Else, it will be $this->db->where_in(). If I didn't do that, I think that it would have put an AND relationship between the two WHERE IN, which would have return no result.

benedmunds:
Ah ok awesome!  Looks good, merging!

avenirer:
Thanks.

sykosoft:
@avenirer Thank you so much. (I had a typo in my #710 where I said groups(), but I meant users() just like you have here). This is a huge help.

avenirer:
Anything that makes the library better, even it is only a proposal, is good for the project. Thank you for taking the time to tell @benedmunds what you needed.

sykosoft:
@benedmunds Looks like the documentation could use an update to reflect this change?

benedmunds:
True, @avenirer could you send a PR with docs updates?

PR #750, opened 2015-03-10
restrict change of name of the admin group

marcelod:


benedmunds:
If this is going to be disallowed in the view it also needs to be disallowed in the controller processing.

marcelod:
OK. Do you have idea where I can do this? If the ion_auth_model/update_group or controller/auth/edit_group is better? And you think you have to get the data for the last id, check the existing name in dads base and if the admin block dai?

avenirer:
The controller is only an example of usage. Important part is the ion_auth model and library.

marcelod:
Now I made a change in the model to validate. Please verify if this right this mode.

benedmunds:
If you return FALSE you also need to set a reason in the errors.  You'll need to add language items as well.

Really I'm not sure if this change is worth the effort in the library.  The users can restrict this in their controllers if needed.

marcelod:
Made changes and added items of languages.

benedmunds:
So I'm really still on the fence about whether or not this should be a library feature...  To me it makes sense for the library to be flexible and users can restrict this in their controllers.

Anyone else have thoughts on this?  Want to try to convince me otherwise?

marcelod:
I understand, but since it blocks areas only to admin group members should block change this name.

avenirer:
I actually like this idea. It is not a library feature but a safety net. It won't allow users to break the library by changing the name of the admins group. :+1:

benedmunds:
Merged.  Thanks for the input.

marcelod:
Thanks for your help.

Issue #510, opened 2013-10-21
Switching between username/email as identity


nathanpitman:
I have a project where I would like to allow users to log-in with a username in one part of the app and have other users elsewhere in the app log-in with email address - I'm guessing that this might be as simple as switching the config var on the fly before I call register or log-in etc

I wondered if this was something that had been attempted by anyone else before and whether there are any immediate pitfalls I should be aware of...

paulcanning:
I have tried this implementation and it's not working for me.

Code:

if($username):
                $forgotten = $this->ion_auth->forgotten_password($username);
            else:
                $this->ion_auth_model->identity_column = 'email';
                $forgotten = $this->ion_auth->forgotten_password($email);
            endif;

When trying to use an email, I just get an error saying "Unable to Reset Password"

paulcanning:
I fixed the problem, by modifying Ion_auth.php, line 144

$identifier = $this->ion_auth_model->identity_column; // new line
$user = $this->where($identifier, $identity)->users()->row();  // changed

benedmunds:
You're reference doesn't seem to match the current codebase.  Can you send a PR against current 2 branch?

paulcanning:
Hmm, I'll have the check what version I'm using, it's been a while! 

Can you provide the reference code from the current codebase just so I can see the difference? 

I'll try do a PR in work tomorrow

benedmunds:
https://github.com/benedmunds/CodeIgniter-Ion-Auth/blob/2/libraries/Ion_auth.php#L144

That is line 144 in the current library

paulcanning:
Looks  like I need line 140?

benedmunds:
I dont see a query using username.  Maybe it was fixed since you last updated your local?

paulcanning:
The issue was that it was using the configuration file no matter what. My code replaced the config with what ever the model was set too, meaning it could be overriden easily

benedmunds:
Ah right. Ok. Send a PR when you get a chance and I'll be happy to merge it.

warisaajkal:
Hi, Sir, I'm a learner of PHP I'm using first time Ion Auth user system

Please add username field on create user form if a new user register where put a username and how to check it new user username not exist in database.

Thanks

benedmunds:
The controller is a example. Feel free to modify as needed.

warisaajkal:
Thanks @benedmunds

PR #751, opened 2015-03-12
Changed forgotten_password method to use model identity column

paulcanning:
As mentioned here https://github.com/benedmunds/CodeIgniter-Ion-Auth/issues/510


benedmunds:
Thanks!

PR #754, opened 2015-03-12
"Removing" unneeded ifs...

avenirer:
I don't know if it's worth the change, but this "merges" two ifs in one. Not that important...


benedmunds:
I actually prefer it with two IFs. It's more explicit and easier for beginners to understand.

PR #755, opened 2015-03-14
updates for Turkish language and reset password in controller

ozdemirburak:
I've updated the Turkish language translations, translated the missing ones, and fixed some mistakes.

On the other hand, in the controller, there is no need for logout after successfully reseting the password. In the current controller, as it destroys the session with no need, no message got displayed in the login page as the logout method is being called, so just redirection for the login page with refresh is better for displaying the message to user as if user has successfully reset his password .


benedmunds:
Thanks!

PR #759, opened 2015-04-04
Update userguide

Tjoosten:


benedmunds:
CI v2 uses the v2 style CI docs.  For now I think I'd like to keep it that way since that will be most familiar to the users.

Anyone else have any thoughts on this vs the old style docs?

Tjoosten:
i think this one is better. because you can maintain you documentation in the code

tlagard:
I say leave it this way for now but let everyone know that it will probably change in the future.

avenirer:
Changing the design of the userguide would be a no from me. I actually don't like how the CI3 manual looks. I liked better the v2 one.

Spaces in code would be a yes from me.

Tjoosten:
@avenirer you got a point there

benedmunds:
Closing for now.  We'll go to a new style for Ion Auth v3 though.

PR #761, opened 2015-04-06
Update index.html

sparky672:
Using "register()", the "Additional Data" array can be left blank or left out entirely without generating an error.  It is not "required".


benedmunds:
Thanks!

PR #764, opened 2015-04-14
synch

Tjoosten:


benedmunds:
We decided not to overhaul the docs right?

benedmunds:
send the spaces as a separate PR

PR #771, opened 2015-04-22
Updated dutch language files

DiederikLascaris:
Ben, I have synced the repositories and I think you should be able to merge it now.


benedmunds:
Thanks!

Issue #402, opened 2013-01-29
users() method returns objects where "id" is ambiguous


jamieburchell:
When using the `users()` method and passing in a group ID the following SQL is generated:

```
SELECT DISTINCT *
FROM (`users`)
INNER JOIN `users_groups` ON `users_groups`.`user_m_id`=`users`.`id`
WHERE `users_groups`.`group_m_id` IN (1)
```

This means that when looping through the results, $user->id is actually the primary key of the `users_groups` table, rather than the user's ID.

zawaruddin:
i'm using sql server and try to get a user by user() function and i get error "Ambiguous column name 'id' "

I think there is an error on line 1400 in ion_auth_model
$this->order_by('id', 'desc');    

so I tried to change it
$this->order_by($this->tables['users'].'.id', 'desc');

benedmunds:
Can you send a PR?

PR #778, opened 2015-05-26
unambigous order_by

avenirer:
...as per issue #402 (although I don't understand why do we need to order by id as long as the limit is already set to 1 and we are given an id...)


benedmunds:
Thanks!

PR #779, opened 2015-05-27
Corrected for compatibility with CI v3

marcopineschi:
For compatibility with Codeigniter v3:

// Produces: DROP TABLE table_name
$this->dbforge->drop_table('table_name');

// Produces: DROP TABLE IF EXISTS table_name
$this->dbforge->drop_table('table_name',TRUE);


benedmunds:
Thanks!

PR #786, opened 2015-06-05
Fix default user's password

kenjis:
I couldn't login.


avenirer:
that is a weird explanation...

kenjis:
What do you mean?

benedmunds:
Thanks!

PR #797, opened 2015-06-19
Update index.html

sparky672:
Comments erroneously state:

>  No need for array('1', '2') as user is always set to member by default

This is simply not true.  The new user will only be put into the exact groups specified within the Group array.  The new user is NOT automatically put into the default group, unless the `Group` parameter is omitted entirely.  Tested on 06/19/2015


benedmunds:
This is correct. Thanks!

PR #805, opened 2015-07-02
Update index.html

sparky672:
The description is misleading.  This function can check if any user is in a group, not just the logged in user.


benedmunds:
Thanks

PR #809, opened 2015-07-06
Update Ion_auth.php

sigmoidal:
Fixed one issue with identity not properly set to the user configured field in when checking if the user is logged in.

Removed the associative array used in unset_userdata, as this is no longer supported in CI 3.x

Please bear in mind you may need to use the old associative array notation in CI 2.x ? (I do not use it, so I cannot test).

### DIFF
--- a/libraries/Ion_auth.php
+++ b/libraries/Ion_auth.php
@@ -386,8 +386,15 @@ class Ion_auth
 
 		$identity = $this->config->item('identity', 'ion_auth');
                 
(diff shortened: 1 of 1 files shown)


benedmunds:
Can you just add a check for the CI version to keep backwards compatibility to CI2?

sigmoidal:
will these be merged?

benedmunds:
Looks good, thanks!

sigmoidal:
thank you too!

Issue #807, opened 2015-07-06
logout() does not work on CI 3.x


sigmoidal:
The logout function in libraries/Ion_auth.php does not work properly in CI 3.x because on line 387, the associative array notation of unset_userdata is used, but this is deprecated:

```
$this->session->unset_userdata( array($identity => '', 'id' => '', 'user_id' => '') );
```

For CI 3.x this should be changed to:
```
$this->session->unset_userdata( array($identity, 'id', 'user_id') );
```

In addition, the userdata functions have been deprecated and therefore it would be nice to update the code for CI3.x eventually.

thanks.

benedmunds:
Could you send a PR please?

sigmoidal:
This was also submitted as a PR some days ago.

PR #814, opened 2015-07-22
fix for #807

Tjoosten:
@benedmunds it think this bug is caught


benedmunds:
Is this just whitespace?

Tjoosten:
Urgh my bad

Issue #819, opened 2015-07-29
A bug in the logged_in() method


Glowieh:
I found a bug in the library function logged_in() while setting up a new CodeIgniter+Ion_auth app with a non-default identity column name.

It's retrieving the user-configured identity column name from session userdata (currently lines 434 and 436 in libraries/Ion_auth.php) while it should be retrieving plain $this->session->userdata('identity'); since the set_session function in the model is saving the relevant data in 'identity', not in the user-configured identity column name (currently line 1744 in models/ion_auth_model.php).

I hope this is a good place to contribute a quick bug report like this one with a fix attached.

benedmunds:
Could you send a PR with the fix?

Glowieh:
Sure, I just didn't know if it was appropriate with such a small fix. I'll do it in a moment.

avenirer:
It's not a bug. In the logged_in() method is retrieving the identity column name from the config in order to retrieve the session data corelated to that identity column. I thouls not be retrieving plain $this->session->userdata('identity') as this would simply allow a user to set anything as identity by just changing the cookies (as per CI version 2.x). The set_session function (method actually) is not saving the relevant data in "identity". In identity is just saving the identity column name.

avenirer:
How did you reach the conclusion that this is a bug? Did the library not work correctly?

Glowieh:
Yeah, the library didn't work as I expected it to. I'll explain what I did from the start so you can point out where I made the misstep since I'm still confused.

In this new CI+Ion_auth app I'm working on I created a replacement database table for the default users table and added some new fields to it that I'm going to need. Among these is the field pinnumber that I'll be using as the login field and configured it as such in ion_auth's config file.

I started testing whether my non-standard database scheme worked and noticed that yes, I can login successfully, but the logged_in() method is returning false. I started debugging. This new pinnumber field isn't one of the default ion_auth fields obviously, so I noticed that the logged_in() method is trying to find 'pinnumber' from userdata, here:

$identity = $this->config->item('identity', 'ion_auth');
return (bool) $this->session->userdata($identity);

It can't find pinnumber from there, because of this, I concluded:

$session_data = array(
    'identity'             => $user->{$this->identity_column},
    'username'             => $user->username,
    'email'                => $user->email,
    'user_id'              => $user->id, //everyone likes to overwrite id so we'll use user_id
    'old_last_login'       => $user->last_login
);
$this->session->set_userdata($session_data);

[...]

PR #820, opened 2015-07-29
Fixed a bug in logged_in() regarding the userdata index

Glowieh:
See issue #819.


benedmunds:
This was correct previously

PR #821, opened 2015-07-29
Proposal to use any identity column and not username

avenirer:
This will allow users to use any field as identity column, and not only "username" (As per issue #819 ).


avenirer:
If @benedmunds will merge this, I will also update the language files in english, and the controller (maybe also the user guide...)

benedmunds:
This is a great change.  Can you update the docs and lang files in this same PR?

avenirer:
@benedmunds I sure hope I didn't miss anything...

benedmunds:
THANK YOU!!!

PR #825, opened 2015-08-01
Update auth.php

ft23:
anybody forgot password :)


benedmunds:
Does the field name in the form match this? I think this was intentional due to the field name being email if I remember correctly.

Might be best to just switch it all to Identity?

Tjoosten:
@ft23 can u plz fix the merge issues?

CC @benedmunds

avenirer:
@benedmunds I don't think this pull request is needed anymore. As I remember, everything works ok with the library

benedmunds:
Thanks!

PR #837, opened 2015-08-12
Update is_admin to is_group

icarlosmendez:
I have modified the is_admin method to allow for the conditional validation of any defined group by passing the group name as the first argument to the method. This is much more flexible and allows developers to add and validate membership in various groups using the same method. See code below. This will require the modification of all instances of 'is_admin' which I believe is just under 20 across the library. See code below and consider the possibilities.

The method:
        /**
	 * is_group
	 *
	 * @return bool
	 * @author Ben Edmunds & Carlos Mendez
	 **/

	public function is_group($g, $id=false)
	{
		$this->ion_auth_model->trigger_events('is_group');

		$g = $this->config->item($g.'_group', 'ion_auth');

		return $this->in_group($g, $id);
	}

Usage:
if (!$this->ion_auth->is_group('admin')) {
     echo "Sorry old chap, you're not an admin!';
}

Regards,
icarlosmendez


icarlosmendez:
Gentlemen,
Obviously I have no intention of merging all these commits but this is my first pull request and I guess I have something to learn about submitting them. Sorry if I messed anything up.
Regards,
icarlosmendez

avenirer:
Why would you change that?...

benedmunds:
The issue is that you set the PR to merge against "master" and you probably meant to merge against "2".

icarlosmendez:
Ben,
Yeah, I realize I didn't do the submission properly and for that I
apologize, but do you think my suggestion has merit? I will resubmit if you
do but maybe I should read up on submitting pull requests.

I will always use the new functionality as I believe it effectively extends
that method allowing it to be used for any number of groups.

Thanks for your consideration.
Regards,
Carlos

benedmunds:
@icarlosmendez I'm not sure what you did exactly since there are so many changes here.  Send a new PR to the 2 branch and I'll review.

PR #838, opened 2015-08-12
update is_admin to is_group

icarlosmendez:
Ben,
I think I figured out how to do the pull request properly this time. 
Have a look and see what you think.
Thanks again for your consideration.
Regards,
Carlos

### DIFF
--- a/controllers/auth.php
+++ b/controllers/auth.php
@@ -23,7 +23,7 @@ class Auth extends CI_Controller {
 			// redirect them to the login page
 			redirect('auth/login', 'refresh');
 		}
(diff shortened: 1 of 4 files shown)


benedmunds:
There's already an in_group() function that does this and a bit more.  is_admin is just a helper function to make a cleaner API.  Thanks though!

icarlosmendez:
Ben,
Quick question. I seem to be limited to 5 groups currently and would like to change that. I thought I remembered seeing a config option for that somewhere but now I can't seem to find it to change it. Did I dream that or is there an option I can mod?

Thanks for your time.
Regards,
Carlos

benedmunds:
There aren't any restrictions on the number of groups.

PR #839, opened 2015-08-12
removed redundant code

avenirer:
...as a FALSE is already returned in line 882.


benedmunds:
I'm not following.. More details?

avenirer:
As you can see from the code, there is an if statement at line 882 that checks for a duplicate identity. If there is one, a FALSE will be returned, so it won't get to line 900. Or did I miss something? Maybe an alternate way would be to set in the config an option for the library to look for an alternate identity if the identity wanted is already taken.

benedmunds:
Ahhh I see.  Thanks!

PR #841, opened 2015-08-15
Auth _render_page function argument changes

msadhrao:


benedmunds:
Thanks

msadhrao:
This was my first pull request on Github :P Sorry for the comment in code at _render_page function definition - "//I think this makes more sense". You can remove the comment.

PR #842, opened 2015-08-17
Input post request name change

BrickGreen:
The name of the post request was changed to retrieve the form input of
the forgotten password view.


benedmunds:
Thanks!

PR #847, opened 2015-08-31
Changes to align with "identity" column

avenirer:
This and the following changes are required in order to accommodate with the "identity" column


avenirer:
This is for the creation of users...

benedmunds:
Looks good. Can you add the new var to the rest of the language files as well, just make the text english as well.

avenirer:
The translation part is by far the hardest one...

benedmunds:
Ha ikr!  Thanks!!!

PR #860, opened 2015-10-08
get_attempt_num() to return results by ip address AND identity...

nplaviola:
when track_login_ip_address is set to TRUE. This corrects an issue where get_attempt_num() would count any login from an ip address, regardless of user, and limit the number of attempts at logging in to less than specified in the config table


benedmunds:
Thanks!

PR #870, opened 2015-10-28
set username default as null

avenirer:
when using another field as identity and not username, the username field can also be null.


benedmunds:
Thanks!

PR #873, opened 2015-10-29
more explaining on register() method.

avenirer:
This is to make sure the users understand that they must put the identity column, even if that is an email.


benedmunds:
Thanks

Issue #881, opened 2015-11-07
failed reset password when change identity type


dexterns88:
When change $config['identity']  to use username 
$config['identity'] = 'username';

Reset password doesn't work , after submit form got message "The Identity field is required".

benedmunds:
Might need a little more help troubleshooting.  Are you passing the username on the reset pass form?

dexterns88:
yes , but they return error " The Identity field is required."

In 'form_validation' you checking email or identity field , but every time field name is 'email', We need to create form field name according to selected authentication type same as when validating form.

```php
if ($this->config->item('identity', 'ion_auth') != 'email') {
      $this->form_validation->set_rules('identity', $this->lang->line('forgot_password_identity_label'), 'required');
} else {
      $this->form_validation->set_rules('email', $this->lang->line('forgot_password_validation_email_label'), 'required|valid_email');
}

if ($this->form_validation->run() == false) {
      // setup the input
      $this->data['email'] = array('name' => 'email',
        'id' => 'email',
        'class' => 'form-control'
);
```

dexterns88:
I made some change in controller forgot_password(), and after this change everything work fine when change auth type to username.

```php
 if ($this->form_validation->run() == false) {
      // setup the input
     // Create field depending on type of authentication
      if ($this->config->item('identity', 'ion_auth') != 'email') {
        $this->data['email'] = array('name' => 'identity',
          'id' => 'identity',
          'class' => 'form-control'
        );
      } else {
        $this->data['email'] = array('name' => 'email',
          'id' => 'email',
          'class' => 'form-control'
        );
      }

// rest of code 
```

```php
} else {
      $identity_column = $this->config->item('identity', 'ion_auth');
      
     // same here
      if ($this->config->item('identity', 'ion_auth') != 'email') {
        $identity = $this->ion_auth->where($identity_column, $this->input->post('identity'))->users()->row();
      } else {
        $identity = $this->ion_auth->where($identity_column, $this->input->post('email'))->users()->row();
      }
//rest of code

```

dexterns88:
This issue is in Ion Auth 2

benedmunds:
Can you send a PR with the change?

dexterns88:
This is my changed forgot_password() function ,


```php
function forgot_password()
  {
    // setting validation rules by checking wheather identity is username or email
    if ($this->config->item('identity', 'ion_auth') != 'email') {
[...]

PR #883, opened 2015-11-09
resolving identity/email bug

avenirer:
as per issue #881


benedmunds:
This looks great, thank you!

dexterns88:
You only change email to identity but if user chose email in config probably will not work?

avenirer:
@dexterns88 actually, i only renamed the form input in the view so that it can work with email or any other identity column. I wrote directly, didn't verify. But if you can verify it would be great. Thank you.

PR #889, opened 2015-11-21
Update ion_auth_model.php allow additional data to override config when register

eakkew:
allow additional data to override config (eg. we need some user to inactive when register, by default users always active when created)

### DIFF
--- a/models/ion_auth_model.php
+++ b/models/ion_auth_model.php
@@ -920,7 +920,7 @@ class Ion_auth_model extends CI_Model
 
 		// filter out any data passed that doesnt have a matching column in the users table
 		// and merge the set user data and the additional data
-		$user_data = array_merge($this->_filter_data($this->tables['users'], $additional_data), $data);
+		$user_data = array_merge($data, $this->_filter_data($this->tables['users'], $additional_data));
 
 		$this->trigger_events('extra_set');
 


avenirer:
I don't understand what you are saying...

benedmunds:
Me either.  This isn't the config...

eakkew:
Sorry for my bad english.

In this code, line# 913 you load config into $data array.
In line# 923 when we use array_merge($arr1, $arr2), the key of array is same the left most is persist  (http://php.net/manual/en/function.array-merge.php).

So, if i set my config `manual_activation` to FALSE, all my register will always auto activate. If i need to activate manually in some case, I cannot; due to line line 923. 

In my code, If I give active=FALSE, I will able to override the manual_activation if in need.

e.g. 
// in my config
$config['manual_activation']          = FALSE;               // Manual Activation for registration

// in my controller
      $additional_data = array(
        'first_name'  => $this->input->post('first_name'),
        'last_name'   => $this->input->post('last_name'),
        'phone'       => $this->input->post('phone'),
        'active'      => FALSE,
      );

      $user_id = $this->ion_auth->register($username, $password, $email, $additional_data, array(3));

avenirer:
That is interesting. I would +1 that. What @eakkew says is that, because of the order of the merging, additional data is overwritten by the configuration data (in this case the manual activation thing).

benedmunds:
CodeIgniter offers a way to override config variables, so I would recommend using that functionality to temporarily change config settings.  See the docs here: https://codeigniter.com/user_guide/libraries/config.html?highlight=config#setting-a-config-item

kapilgtm:
Ion_Auth is compatible with ajax or not !! because its working with the simple login but when using ajax its not working !! pls provide me soltn

kapilgtm:
public function login(){
         
		$this->data['title'] = "Login";

		//validate form input
		$this->form_validation->set_rules('identity', 'Identity', 'required');
		$this->form_validation->set_rules('password', 'Password', 'required');

		if ($this->form_validation->run() == true)
		{
			// check to see if the user is logging in
			// check for "remember me"
			$remember = (bool) $this->input->post('remember');

			if ($this->ion_auth->login($this->input->post('identity'), $this->input->post('password'), $remember))
			{
		                echo "success";
			}
			else
			{
				echo "fail";
                	}
		}
		else
		{
			echo "fails";
                    
		}
} 

This function always return fail when using ajax  .
But with out ajax its working fine. pls guide me

avenirer:
What do you mean by "Ion_Auth is compatible with ajax or not"? Are you for real?

kapilgtm:
Nothing I have resolved the issue . Thanks for the concern avenirer !!

PR #894, opened 2015-12-10
What do i do next?

Mblockmus13:


benedmunds:
What are you doing?

Mblockmus13:
simple fixes

benedmunds:
This doesn't fix anything.  Why would this be included in the library?  What is Prompt?

avenirer:
Where is the institution of banning when you need it...

PR #917, opened 2016-02-22
Added Get Group Users

steve-goodwin:
Handy for getting a list of users who are apart of a specific group


nathanpitman:
:+1:

avenirer:
For list of users in a group you already have the users() method. Or am I wrong?

steve-goodwin:
@avenirer Reviewing the docs, your actually right! My bad, I always use this method, but thought it just brought back all users in the current table.

I've not checked the docs for a while, so maybe an update sneaked in!

PR #922, opened 2016-03-10
we remove the posibility of setting 0 as user_id

avenirer:
I don't know how this helps, but is good to make sure if no id was passed or not.


benedmunds:
Thanks

Issue #949, opened 2016-05-14
`ORDER BY id DESC` instead of `Select MAX(time)`


zilzurna:
Hi, i use this library for first time.

I think there is a performance issue in get_last_attempt_time function.

This is will be better instead of current version.

`$this->db->select('time');
if ($this->config->item('track_login_ip_address', 'ion_auth')) $this->db->where('ip_address', $ip_address);
else if (strlen($identity) > 0) $this->db->or_where('login', $identity);
$this->db->order_by('id', 'desc');
$qres = $this->db->get($this->tables['login_attempts'], 1);`

Current version is, 
`$this->db->select_max('time');
if ($this->config->item('track_login_ip_address', 'ion_auth')) $this->db->where('ip_address', $ip_address);
else if (strlen($identity) > 0) $this->db->or_where('login', $identity);
$qres = $this->db->get($this->tables['login_attempts'], 1);`

What do you think?

benedmunds:
Can you provide benchmarks or explains explaining the reasoning and advantage for this change?

zilzurna:
I have no benchmark yet but in logic the max function is checking all rows, cause date column has not index like id column. If we ordering id column by desc then first one will be last attempt time for each user all the time.

benedmunds:
Yea that makes sense.  Can you send a PR?

zilzurna:
i succeed it. oh at last :|

benedmunds:
@zilzurna looks like you PR'd it against your own project.  You'll need to set the PR against this repo for it to be merged in.

benedmunds:
Closing due to inactivity. Please comment if this is stil an issue.

PR #951, opened 2016-05-15
Update Ion_auth_model.php

zilzurna:
#949


benedmunds:
Should this also have a limit of 1?

zilzurna:
@benedmunds  It already has limit 1.

$this->db->get($this->tables['login_attempts'], **1**)

benedmunds:
Yup you're right.  Merging.  Thanks!!

zilzurna:
Thank you!

PR #964, opened 2016-07-07
Protecting the whole controller at once.

retiredment:
Protecting the whole controller once and white list the few pages left that don't need this (Removed all addition logged_in() checks).
Also we're checking in reset_password() if the code is provided, lets do the same for activate method's id.


lloricode:
thanks for this

lloricode (approved):

benedmunds:
Closing. The controller code is just an example and I feel that the current way is more explicit and easier to understand for new users

PR #969, opened 2016-07-29
Change DB structure for more flexible additionals data

AsoStrife:
Ehy Ben, I want suggest you a different approach to save additionals data for users. Check out this E-R
[E-R Image Screenshot](http://i67.tinypic.com/9k4ara.jpg)
With a different table for additional fields, we can save more and more data, even different data for different account. This is very useful if we need to manage different type of account. 

Thanks!


benedmunds:
Hey @AsoStrife 

Thanks for taking the time to send this.  Ion Auth v1 actually followed this pattern, we called it a meta data table.  They was replaced with what we have now simply because the meta table was confusing to many and there is no real need to have the additional overhead of a related table when we can store information in the table that the data belongs to.  I can definitely understand that this pattern may be helpful with a large application though and commend you for getting it working for your needs.

So thanks so much but going to turn down this PR.

AsoStrife:
Thanks for your answer :)

PR #978, opened 2016-09-13
Fixed Ion_auth_model for username input

DavidSili:
Before the update, if the setting for identity in ion_auth config was "email", when creating a new user a username field would be left blank. This fixes the issue for both ways to identify users (both "email" and "username").

This solution was given by Aleix.

### DIFF
--- a/models/Ion_auth_model.php
+++ b/models/Ion_auth_model.php
@@ -899,6 +899,7 @@ class Ion_auth_model extends CI_Model
 		// Users table.
 		$data = array(
 		    $this->identity_column   => $identity,
+		    'username'   => $identity,
 		    'password'   => $password,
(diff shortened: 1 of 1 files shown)


benedmunds:
Thanks!

DavidSili:
Thank You :) I was searching for some time for such code to handle
user-controll. This was just a tiny contribution to the community.
I do plan in following months to extend it to include at least some of the
options:
- to allow changing emails (only if identity is set to username)
- protection against session hijacking
- captcha
- optional 2FA

benedmunds:
That sounds great.  I probably wouldnt want to include most of that in the core library, but if you could do it with hooks that would be awesome and I'd be happy to link to it from this libraries readme.

PR #990, opened 2016-10-21
Add get_all_attempts() function to model

Indigo744:
I needed this function from a manager point of view (my client wanted to be able to look at all attempts in one place).

This function will return an array populated with all useful information regarding attemps (see list in comments). I used the same key names as existing function:
- 'attempts_num' refers to function `get_attempts_num()`
- 'last_attempt_time' refers to function `get_last_attempt_time()`
- 'is_time_locked_out' refers to function `is_time_locked_out()`

### DIFF
--- a/models/Ion_auth_model.php
+++ b/models/Ion_auth_model.php
@@ -1129,6 +1129,53 @@ class Ion_auth_model extends CI_Model
 		}
 		return FALSE;
 	}
+    
+	/**
+     * Get all attemps from all users
+     * 
+     * Format of return array is:
+     *      [0] => array("login" => string
+     *                   "ip_address" => string  // only if track_login_ip_address is enabled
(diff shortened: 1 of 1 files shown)


benedmunds:
I'm going to close this PR without merging.  Not because there are any issues with it, just because i don't believe it will have wide spread enough need to justify adding it to the core API.  I will be happy to point people here if anyone needs it going forward.  Thanks!

Indigo744:
Aww.. Ok, I understand..
Quick question git/github related: if I ever delete this branch, will this stay as is?

benedmunds:
I'm actually not sure..

Indigo744:
I just deleted the branch, and the commit and modifed files are still visible. Good!

benedmunds:
👍 👍 👍   Thanks!

PR #991, opened 2016-10-21
Check post var against NULL instead of FALSE

Indigo744:
`$this->input->post()` now returns NULL instead of FALSE when the required items don’t exist.

See for reference: http://www.codeigniter.com/user_guide/installation/upgrade_300.html#step-10-many-functions-now-return-null-instead-of-false-on-missing-items

### DIFF
--- a/controllers/Auth.php
+++ b/controllers/Auth.php
@@ -795,7 +795,7 @@ class Auth extends CI_Controller {
 
 	public function _valid_csrf_nonce()
 	{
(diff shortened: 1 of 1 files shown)


benedmunds:
Could you check both so we dont lose CI 2 compatibility?

Indigo744:
Done, what do you think ? 

I simply check `if($csrfkey)`, it will return FALSE for FALSE or NULL or empty string.

Indigo744:
Note that I can squash the commit if you want.

benedmunds:
Looks good, thanks!

Issue #1001, opened 2016-12-05
problem on remember login


italoc:
i have a problem with remember login, the cookie always expire when 'sess_expiration' is exceeded and the users need to be relogged,  i need that user is always remember if i check REMEMBER LOGIN, and is loggeout when close the browser if REMEMBER LOGIN is not checked
i'm using C.I. 3.1.2 and latest version of ion_auth, with older version of C.I. 2.2.4 all work great but not with latest version

this is my setting

`$config['sess_driver'] = 'database';
$config['sess_cookie_name'] = 'aaw_ci_session';
$config['sess_expiration'] = 2592000;
$config['sess_save_path'] = 'ci_sessions_test';
$config['sess_match_ip'] = FALSE;
$config['sess_time_to_update'] = 300;
$config['sess_regenerate_destroy'] = FALSE;`

`$config['remember_users']             = TRUE;               
$config['user_expire']                = 0;                  
$config['user_extend_on_login']       = TRUE;            
$config['track_login_attempts']       = TRUE;          
$config['track_login_ip_address']     = TRUE;           
$config['maximum_login_attempts']     = 3;                
$config['lockout_time']               = 600;            
$config['forgot_password_expiration'] = 0;`

can you help me?

avenirer:
Did you try  ```$config['sess_expiration'] = 0```?

italoc:
yes if i set ` $config['sess_expiration'] = 0` login is never remember every time i need to login

avenirer:
Can you show us the login controller?

italoc:
it's the standard login function, untouched from latest version of ion_auth

`	
public function login()
	{
		$this->data['title'] = $this->lang->line('login_heading');
		//validate form input
		$this->form_validation->set_rules('identity', str_replace(':', '', $this->lang->line('login_identity_label')), 'required');
		$this->form_validation->set_rules('password', str_replace(':', '', $this->lang->line('login_password_label')), 'required');

		if ($this->form_validation->run() == true)
		{
			// check to see if the user is logging in
			// check for "remember me"
			$remember = (bool) $this->input->post('remember');
                        
			if ($this->ion_auth->login($this->input->post('identity'), $this->input->post('password'), $remember))
			{
				//if the login is successful
				//redirect them back to the home page
				$this->session->set_flashdata('message', $this->ion_auth->messages());
				redirect('/', 'refresh');
			}
			else
			{
				// if the login was un-successful
[...]

PR #1002, opened 2016-12-07
"remember me" doesn't work correctly

avenirer:
If identity column is the email then, when the email is saved in the cookie the "@" symbol is "%40" and when the database is queried, it will return false.
As per issue #1001. @italoc maybe you should try this change to see if it works before it is merged.


benedmunds (approved):

benedmunds:
Merged.  Thanks!

PR #1013, opened 2016-12-21
Fixed 'login attempts' functionality. 

wadealer:
Methods 'get_attempts_num' and 'get_last_attempt_time' had different logic. This is fixed.
We should check both IP address and itentity to avoid bruteforce different logins from one IP as well as avoid bruteforce one login from different IPs.

### DIFF
--- a/models/Ion_auth_model.php
+++ b/models/Ion_auth_model.php
@@ -1047,19 +1047,18 @@ class Ion_auth_model extends CI_Model
 	 * @param	string $identity
 	 * @return	int
 	 */
-	public function get_attempts_num($identity)
-	{
-        if ($this->config->item('track_login_attempts', 'ion_auth')) {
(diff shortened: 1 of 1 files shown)


benedmunds:
I previously didn't prevent from one IP hitting multiple users since corporate users will often share an IP.  This could lead to quickly not being able to login under normal conditions.

wadealer:
Yes, this is the problem. But this problem can be solved by properly setting the other parameters (increasing maximum_login_attempts, decseasing lockout_time).  Also, if the majority of users - corporate users, you can disable the option track_login_ip_address.

benedmunds:
You won't always know if you have corporate users.  For example, a lot of SASS apps will have both consumers and businesses as clients.  I'd rather not merge this in as is when it will likely cause issues for existing users.  If you add in the check for both I'll be happy to merge it.

Issue #1024, opened 2017-01-29
Codeigniter Ion Auth configuration and Email Activation


salihkulangara:
I have seen a lot of questions about email activation related, but nothing resolved my issue. To simply put, I have asked this same question on SO, but no luck. here is the issue

 I need email activation for new registrations [I am not using SMTP, there is no email config file in my config folder], So I made changes in ion_auth.php in config file as follows::

```
$config['email_activation']           = TRUE; 
$config['manual_activation']          = FALSE; 
$config['use_ci_email'] = TRUE;
```
as far as I know about this library (I am new to CodeIgniter and Ion Auth), that's it. But unfortunately these configurations are not taking effect. When I var dump the following items inside constructor of ion_auth.php library file returns null

```
$this->config->item('use_ci_email','ion_auth') or
$this->config->item('email_activation','ion_auth')
```
And other config items like 'admin_email' has no issues, I can fetch the value. So, what is wrong with my config file, I am using very basic file provided by the library repo.

When I set the values, inside the constructor of library file as below, it works.

```
$this->config->set_item('use_ci_email', TRUE);
$this->config->set_item('email_activation', TRUE);
```
Is there anything wrong I am doing to get this behavior. Do I need a emial.php config file with SMTP and other details, even if I am not using SMTP[yet I have tried that also].

I am not getting any error as you might know. What happens is By registering registration happens straight and activates new user by default.

benedmunds:
Post your entire config file please.

Also ensure you aren't autoloading the config file or loading it anywhere outside of Ion Auth.

salihkulangara:
Thanks for your replay, as I said I haven't changed much in the config file, that I got from the git repo, just changed the admin email, and above said config Items. Also I made sure that I am not auto loading IonAuth config file anywhere out side IonAuth Library file.  For making that sure I am using just the provided ionAuth controller file Auth.php and corresponding view files.
Here is the code I have in my config file:::

```
<?php  if ( ! defined('BASEPATH')) exit('No direct script access allowed');
/**
* Name:  Ion Auth
*
* Version: 2.5.2
*
* Author: Ben Edmunds
*		  ben.edmunds@gmail.com
*         @benedmunds
*
* Added Awesomeness: Phil Sturgeon
*
* Location: http://github.com/benedmunds/CodeIgniter-Ion-Auth
*
* Created:  10.01.2009
*
[...]

PR #1025, opened 2017-01-31
how do we load the configuration?

avenirer:
Don't ask me about this @benedmunds... Maybe in connection with issue #1024

But I am amazed about how it worked up until now. Am I missing something?


avenirer:
Oh... I see now... https://www.codeigniter.com/user_guide/libraries/loader.html#CI_Loader::config
But being an alias maybe is not as good as the main way of loading the configuration file.

benedmunds:
Weird, merged.  Thanks!

PR #1031, opened 2017-02-03
Add OTP to edit_user function

michaelsmoody:
This patch allows for the edit_user function to enable/disable OTP.


lloricode:
maybe you need only one pull request in your two commits.? for easy test?

PR #1035, opened 2017-02-05
remove dynamically identity in requesting forgot password code

lloricode:
only user_email will parameter in two functions

```
$this->ion_auth->forgotten_passwor($user_email);
$this->ion_auth_model->forgotten_password($user_email);
```

### DIFF
--- a/config/ion_auth.php
+++ b/config/ion_auth.php
@@ -42,12 +42,6 @@ $config['join']['users']  = 'user_id';
 $config['join']['groups'] = 'group_id';
 
 /*
- | table column for email in user to use in requesting forgot password code
- |
- */
-$config['column_email_forgot_password']  = 'email';
-
-/*
  | -------------------------------------------------------------------------
  | Hash Method (sha1 or bcrypt)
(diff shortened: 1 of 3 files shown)


lloricode:
@avenirer i hope is this ok,

avenirer:
@lloricode @benedmunds I really don't understand the reason for doing this change. This is not doing anything new. It is only adding a new unnecessary config item. Instead of letting the library take the email address from the user row, it takes the configuration column for email to retrieve it.

lloricode:
@avenirer  hmm much better.  in two file i just set email even identity is username

avenirer:
@lloricode But that is just what I tell you. The model is already taking the email field from the user result. There is no need for these changes.

lloricode:
@avenirer  I understand. maybe my usage/flow is the problem, by the way thanks, now i have idea to use to solve my issue.

lloricode:
hmm im still confusing

benedmunds:
I'm not following either. The user record is queried by the identity and then the email is used from the user record

PR #1036, opened 2017-02-07
More specific message for unsuccessful send

avenirer:
This should be a more specific error description. @benedmunds maybe you can find a better message.


benedmunds:
Good idea, thanks!

Issue #1039, opened 2017-02-11
Error when using like-method on model


tookig:
The method `like `in `ion_auth_model `seems to be using a format for the like-array that is not supported by CodeIgniter:
```
$like = array($like => array(
      'value'    => $value,
     'position' => $position
));
```
I can't find anything in the CI docs about adding the position when using an array with `like`. 
Probably easiest just to remove the position functionality:

```
	public function like($like, $value = NULL)
	{
		$this->trigger_events('like');

                if (!is_array($like)) {
                     $like = array($like => $value);
                }

		array_push($this->_ion_like, $like);

		return $this;
	}
```

Otherwise the array needs to be mapped to single like-calls whenever it is used.

benedmunds:
What version of ion auth are you using?

I just checked the v3 code and it looks like you're right.  Can you send a PR for this change?

benedmunds:
For reference: https://github.com/bcit-ci/CodeIgniter/blob/develop/system/database/DB_query_builder.php#L955

PR #1041, opened 2017-02-12
Fix error when using like-method on ion_auth_model, #1039

tookig:
Fixes the issue, and keeps the position-functionality, however it also changes the format of the `_ion_like` array (is that acceptable?).


benedmunds:
Thanks!!!

PR #1042, opened 2017-02-12
Fix desactivate function when user is'nt connected

celianvdb:
$this->user()->row()->id is not defined if user is'nt connected.


benedmunds:
You shouldnt ever call decactivate if you arent logged in.  Can you describe a use case where this would be needed?

celianvdb:
In my signup page when i want to add user with register(), if i enable email_activation it will call desactivate and at this moment user is'nt logged.

benedmunds:
Ah that makes sense, merging.  Thanks!

Issue #1045, opened 2017-02-14
Destroy a session when user is deleted from database


namanbansal45:
I deleted user from database who is currently logged in. how session will be invalidated here, so if user is deleted he must be redirected to login page?

benedmunds:
They will remain logged in unless you check against the database.

PR #1046, opened 2017-02-14
recheck session against database after some defined time

avenirer:
as per issue #1045, it seems like a good idea... It is a proposal, so it can be repelled.

### DIFF
--- a/config/ion_auth.php
+++ b/config/ion_auth.php
@@ -99,7 +99,10 @@ $config['track_login_ip_address']     = TRUE;                // Track login atte
 $config['maximum_login_attempts']     = 3;                   // The maximum number of failed login attempts.
 $config['lockout_time']               = 600;                 // The number of seconds to lockout an account due to exceeded attempts
 $config['forgot_password_expiration'] = 0;                   // The number of milliseconds after which a forgot password request will expire. If set to 0, forgot password requests will not expire.
-$config['recheck_timer']              = 0;                  // The number of seconds after which the session is checked again against database to see if the user still exists and is active. Leave 0 if you don't want session recheck.
+$config['recheck_timer']              = 0;                   /* The number of seconds after which the session is checked again against database to see if the user still exists and is active.
+							           Leave 0 if you don't want session recheck. if you really think you need to recheck the session against database, we would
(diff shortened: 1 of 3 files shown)


benedmunds:
I like this idea!

benedmunds:
Just let me know when the PR is ready for review, looks like it's still in active dev atm

avenirer:
@benedmunds ready...

avenirer:
it should be backwards compatible...

benedmunds:
Walk me through.  The plan is to recheck their session if they logged in longer than X seconds ago?

avenirer:
1. check if the user has set up a recheck_timer
2. if it did, when method logged_in() is called, the recheck_session() method from ion_auth_model is called
3. in recheck_session() we check if the current time plus the recheck_timer in seconds is smaller than current time. if that's the case then the user is verified against database. if the user doesn't exist or it is inactive, the session data is unset and the method returns false. Else, the session data 'last_login' is reset to current time() and the method returns true.
4. true/false (what is returned from recheck_session()) is then returned from logged_in().

benedmunds:
Makes sense.  Couple things:

1. How about we change the default in the config to something like ~5 min.  I want to promote a performant default here.

2. What do you think about renaming last_login to something else? Since its more like a last_modified or last_update or something rather than last_login...

avenirer:
1. I would rather leave it to zero and let the users decide (maybe mention "if you really think you need to recheck the session against database, we would recommend a higher value, as this would affect performance" in the comment). If it's set to zero, the session won't be rechecked.
2. that would be a great idea... maybe "last_check"?

benedmunds:
@avenirer 1 - yup that sounds great.  2 - agreed, works for me.

benedmunds:
Oh one other thing.  is there any need to do the full query instead of just $this->user()?

avenirer:
$this->user() would return quite some data, just like the query I've done in there... which is a lot and not needed... So I only asked for id now. It's more than enough to return us one row.

benedmunds:
Looks great, thanks man.  Merging!

lloricode (commented):
more secure :)

PR #1050, opened 2017-02-19
add table for users_last_logins, then also on update column of users

lloricode:
if this is ok, i will also PR for getting all last_login by current user or as parameter of user(s)

### DIFF
--- a/config/ion_auth.php
+++ b/config/ion_auth.php
@@ -31,6 +31,7 @@ $config['tables']['users']           = 'users';
 $config['tables']['groups']          = 'groups';
 $config['tables']['users_groups']    = 'users_groups';
 $config['tables']['login_attempts']  = 'login_attempts';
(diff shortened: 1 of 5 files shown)


avenirer:
Why this?

lloricode:
@avenirer  to see the all last login in all users?. because i want to monitor all last logins

avenirer:
I don't think this in needed. You can do this inside any controller by reading the session variable last login.

lloricode:
@avenirer , but not on session we want to get the last login, we just want to get the all from the past login?

lloricode:
because once user login again, it will override the time() in users row, so saving in other table is the other way,

avenirer:
Yes. On every successful login you can do it in your own flow.

benedmunds:
Closing, this isn't something we'd want to add to the core.

lloricode:
@benedmunds , ok sir

Issue #1052, opened 2017-02-24
Function mcrypt_get_iv_size() is deprecated


Th3AL0:
Hello Guys, I'm getting this error in the forgot_password. Thanks in advance!

![error](https://cloud.githubusercontent.com/assets/5881465/23300462/f5063016-fa85-11e6-9280-24123f12e1dc.PNG)

AlexandreLogist:
@avenirer @benedmunds  Thanks !

Th3AL0:
Thanks for solving this issue!

amitamie-com:
Hello Guys I was getting same above error message but by updating the code which was suggested from @avenirer is solved.
But still I am facing some issues and I am getting below error.

> A PHP Error was encountered
> Severity: 8192
> Message: Function mcrypt_get_iv_size() is deprecated
> Filename: libraries/Encrypt.php
> Line Number: 337
> 
> A PHP Error was encountered
> Severity: 8192
> Message: Function mcrypt_decrypt() is deprecated
> Filename: libraries/Encrypt.php
> Line Number: 346

Please help ! Thanks in Advance !!!

benedmunds:
@amitamie-com is this a duplicate of https://github.com/benedmunds/CodeIgniter-Ion-Auth/issues/1076#issuecomment-297697292 ?

amitamie-com:
**@benedmunds**  Sorry for duplicate post , actually I thought two comments at two different places are better than one , **but guys replied so fast** , for that thank you so much , and in future I will not do any duplicate comment.

benedmunds:
Ok cool, take care!

julio899:
i can solve using the new library encryption recommend.
example:
`
// #Encrypt

        $this->load->library('encryption');
        $to_encrypt="some value";
        $this->encryption->initialize(
                array(
                        'cipher' => 'aes-256',
                        'mode' => 'ctr',
                        'key' => $this->config->config['encryption_key']
                )
        );
        echo $this->encryption->encrypt($to_encrypt);
`

PR #1053, opened 2017-02-24
deprecated function since 7.1

avenirer:
as per issue #1052

### DIFF
--- a/models/Ion_auth_model.php
+++ b/models/Ion_auth_model.php
@@ -375,7 +375,13 @@ class Ion_auth_model extends CI_Model
         $buffer_valid = false;
 
         if (function_exists('mcrypt_create_iv') && !defined('PHALANGER')) {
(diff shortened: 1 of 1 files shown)


benedmunds:
@avenirer  What about doing this instead, to keep the same style of feature checks:


    if (function_exists('random_bytes')) {
      $buffer = random_bytes($raw_salt_len); 
      if ($buffer) {
        $buffer_valid = true;
      }
    }

    if (!$buffer_valid && function_exists('mcrypt_create_iv') && !defined('PHALANGER')) {
         $buffer = mcrypt_create_iv($raw_salt_len, MCRYPT_DEV_URANDOM);     
        if ($buffer) {
            $buffer_valid = true;
        }
    }

avenirer:
You are right. That would look better.

benedmunds:
@avenirer Please review and comment: https://github.com/benedmunds/CodeIgniter-Ion-Auth/pull/1054

PR #1063, opened 2017-04-13
Catalan translation

duub:
Complete the translation into the catalan language files


benedmunds:
Looking good, thank you!

PR #1064, opened 2017-04-13
Some lines missing in English lang file

duub:
I found that in the english file those lines were missing

### DIFF
--- a/language/catalan/auth_lang.php
+++ b/language/catalan/auth_lang.php
@@ -3,7 +3,7 @@
 * Name:  Auth Lang - English
 *
 * Author: Ben Edmunds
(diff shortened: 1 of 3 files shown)


avenirer:
Most likely they are missing because they are no longer in use by the library. The English language files should always be considered the template for the other language files and not the other way around.

duub:
Ok! I didn't know that. I have been translating into catalan, so you are saying that instead of adding into English I should remove it from the catalan file? The english version is the one it is updated?

benedmunds:
@duub yea that's correct, the english version has been updated most likely.

duub:
Ok! then We can discard the PR

PR #1066, opened 2017-04-15
edit only group description

lloricode:
allowing edit all description,but not all name of groups,

the parameter will be is the ID and Description. while the name set to FALSE to ignore update on name.

```php
          //after validation input
	   $group_update = $this->ion_auth->update_group(
                $id, FALSE, $this->input->post('description', TRUE)
         );

```

### DIFF
--- a/models/Ion_auth_model.php
+++ b/models/Ion_auth_model.php
@@ -1958,7 +1958,7 @@ class Ion_auth_model extends CI_Model
 
 		// restrict change of name of the admin group
         $group = $this->db->get_where($this->tables['groups'], array('id' => $group_id))->row();
-        if($this->config->item('admin_group', 'ion_auth') === $group->name && $group_name !== $group->name)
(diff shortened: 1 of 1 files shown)


benedmunds:
I'm not sure I follow, can you explain the need for this and then the solution you've implemented here please.

lloricode:
hi @benedmunds,

admins can modify all description only even `admin`'s description, not include all `name`.

benedmunds:
Why?

lloricode:
@benedmunds , in my case, the description of current user_group's is in top.
screenshot at 2017-04-16 09-26-52

so admin can rename it.
hmm is this ok?

benedmunds:
There's no probably with the admin renaming groups at a library level.

If you need to stop that from happening in your app you should do it in your controller.

lloricode:
@benedmunds hhmm ok sir,

PR #1067, opened 2017-04-15
fix a label and other minor things

duub:
- Fixing the label 'forgot_password_identity_label' in the line 134
- The changes between the lines 129 to 136, except for the 134, are the alignment of the '='
- And delete unused strings that doesn't exists in the english version (from line 145 to 158)


benedmunds (approved):

benedmunds:
Thanks!

PR #1070, opened 2017-04-20
missing in russian lang

lloricode:


benedmunds:
Thanks! Can you translate as well?

lloricode:
@benedmunds ,I've just add a $lang[] in file, because i seen not in there, but the current language is not our

benedmunds:
Ok thanks!

PR #1071, opened 2017-04-20
add missing lang in forgot_password_identity_not_found

lloricode:
I just translate our language FILIPINO, so the other is just English

[here](https://github.com/benedmunds/CodeIgniter-Ion-Auth/blob/2/controllers/Auth.php#L225) the lang tag that not exist.
```php 
        public function forgot_password()
	{
               .....

               if ( $this->config->item('identity', 'ion_auth') != 'email' ){
		     $this->data['identity_label'] = $this->lang->line('forgot_password_identity_label');
		}
		else
		{
	            $this->data['identity_label'] = $this->lang->line('forgot_password_email_identity_label');
	        }
 ```


benedmunds:
Thanks!

Issue #1075, opened 2017-04-22
access trigger with remember 


lloricode:
Good pm, @benedmunds.

I just want to set hook with this. 

```php
	public function set_session($user)
	{

		$this->trigger_events('pre_set_session');

		$session_data = array(
		    'identity'             => $user->{$this->identity_column},
		    $this->identity_column             => $user->{$this->identity_column},
		    'email'                => $user->email,
		    'user_id'              => $user->id, //everyone likes to overwrite id so we'll use user_id
		    'old_last_login'       => $user->last_login,
		    'last_check'           => time(),
		);

		$this->session->set_userdata($session_data);

		$this->trigger_events('post_set_session');

		return TRUE;
	}
```

it works fine though...

but when `remember` is set to `TRUE`  after login user, then reach the `user_expire`

I see that the function I want to trigger is not called

even `pre_set_session` OR `post_set_session`

Thanks for response :)

benedmunds:
Sorry I'm not sure if I'm following exactly. 

So with a remembered user, the pre_set_session event is not triggering?  Can you add debug code to ensure it's getting to that line in the code?

Is the post_login_remembered_user event triggering?

lloricode:
I share my test code



##
just to make sure
**index.php**

`define('DEBUG', 'CI_DEBUG ');`

##
**config/ion_auth.php**
`$config['user_expire']                = 0;   `
##
**config/config.php**
```php
$config['sess_expiration'] = 3;   //3 seconds for testting
  $config['log_threshold'] = 2;
```
##
adding log message
**Ion_auth_model.php**
```php
public function set_session($user)
{

        log_message('debug', DEBUG.'pre_set_session');//<------------------
        
        $this->trigger_events('pre_set_session');

        $session_data = array(
            'identity'             => $user->{$this->identity_column},
            $this->identity_column             => $user->{$this->identity_column},
            'email'                => $user->email,
            'user_id'              => $user->id, //everyone likes to overwrite id so we'll use user_id
            'old_last_login'       => $user->last_login,
            'last_check'           => time(),
        );

        $this->session->set_userdata($session_data);

        $this->trigger_events('post_set_session');
        
        log_message('debug', DEBUG.'post_set_session');//<------------------

        return TRUE;
}
```

##
**controller**
`print_r($this->session->userdata());`


##
**core/MY_Controller.php**
```php
function __construct()
{
        parent::__construct();
[...]

PR #1077, opened 2017-04-27
access trigger with remember login

lloricode:
this can be resolve issue #1075 ,

Ive noticed that checking remember is on `contruct`, then i see it also check  `logged_in` ,

so i decide to combine the function `logged_in` and the checking `remember`.

so `ion auth `can load entire file before check if `remeber` login

### DIFF
--- a/libraries/Ion_auth.php
+++ b/libraries/Ion_auth.php
@@ -439,6 +439,10 @@ class Ion_auth
 
                 $recheck= $this->ion_auth_model->recheck_session();
         
+                if(! is_bool($recheck))//check if the return is bool, else mean the recheck check it that need to log out the user
+                {
+                      return FALSE;  
+                }
                 //auto-login the user if they are remembered
                 if ( ! $recheck && get_cookie($this->config->item('identity_cookie_name', 'ion_auth')) && get_cookie($this->config->item('remember_cookie_name', 'ion_auth')))
 		{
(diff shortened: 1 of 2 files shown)


lloricode:
also i notice that when `recheck` is FALSE,  it will still login in when using `remember` login,
so it will ignore the `recheck` database function with using `remember` login

benedmunds:
Idea of the changes look good.  Do you need that is_bool check though?  From what I see the function always returns a bool...

lloricode:
the return of `recheck_session` has a two of FALSE either `session` is expire OR the user detect `inactive`.

so maybe need to identify what mean of FALSE returned by a function.


to prevent using remember login event the user is inactive.


hmmm im not sure the change maybe it need more enhance. im still confusing here

lloricode:
hmm sir @benedmunds , can you test it? or improve it?

benedmunds:
Hey,

I'm just going to close this PR.  I'm all for making things more efficient but not at the detriment of clarity and backwards compatibility.

1.  I think the new logic for the controller is harder to follow.

2.  I don't want to change the current functionality of the model method because users may be expecting it to work as is and this change has the risk of breaking backwards compatibility.


Thanks so much for the contribution though and take care!

lloricode:
yes the compatibility is important,
thanks @benedmunds  :)

lloricode:
hi @benedmunds , can I PR only fiirst commit ? it should be first load entire library before check the remember.

benedmunds:
Still not sure how I feel about this.  There is nothing forcing the logged_in() method call so it seems like this may still break BC for some users.

@avenirer do you have an opinion here?

avenirer:
I am sorry @lloricode @benedmunds . I would have to look at it in the following days, if you have time to wait :(

lloricode:
@benedmunds maybe @avenirer has much idea, so we can wait :)

PR #1078, opened 2017-04-28
load entire lib b4 check remember

lloricode:
it will give a change to set hook before check remember,

like `post_set_session` that not called because the check remember is on construct, not give a change to set the hook first

this will solve issue #1075 , and also remove some commit in PR #1077

### DIFF
--- a/libraries/Ion_auth.php
+++ b/libraries/Ion_auth.php
@@ -66,13 +66,7 @@ class Ion_auth
 		$this->load->model('ion_auth_model');
 
 		$this->_cache_user_in_group =& $this->ion_auth_model->_cache_user_in_group;
-
-		//auto-login the user if they are remembered
-		if (!$this->logged_in() && get_cookie($this->config->item('identity_cookie_name', 'ion_auth')) && get_cookie($this->config->item('remember_cookie_name', 'ion_auth')))
-		{
-			$this->ion_auth_model->login_remembered_user();
-		}
-
+	
 		$email_config = $this->config->item('email_config', 'ion_auth');
 
 		if ($this->config->item('use_ci_email', 'ion_auth') && isset($email_config) && is_array($email_config))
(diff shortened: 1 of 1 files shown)


avenirer:
This PR seems ok to me, because, just as the corrected code says, the loading of the library doesn't imply a check on the login status. The constructor shouldn't check if the user is logged_in(), nor should log in the user automatically.

benedmunds:
My concern is that it changes the logic flow for backwards compatibility.  So a user wont be logged in (if remembered) until logged_in() is checked...  

I've looked through the code and the examples and cant find any places where this causes an issue though so I'll go ahead and merge it.  Noting this here in case anyone raises concerns in the future, we'll know where to look.

Thanks @lloricode and @avenirer

lloricode:
@benedmunds thanks for merging, and @avenirer thanks for reviewing.

then I have some mention, @avenirer , in this PR I've just not modify the behavior of library, but just moved a existing code in much better place, to make chance to load set hook before trigger some hooks.

**issue:**
then the recheck is ignore if `remembered `user login, even the return of recheck is need user to log out because user logged in is `inactive`

##
this is the code from `recheck_session`
```php

if ($query->num_rows() === 1)
{
       $this->session->set_userdata('last_check',time());
}
else
{
       $this->trigger_events('logout');
       $identity = $this->config->item('identity', 'ion_auth');
       if (substr(CI_VERSION, 0, 1) == '2')
       {
              $this->session->unset_userdata( array($identity => '', 'id' => '', 'user_id' => '') );
       }
       else
       {
              $this->session->unset_userdata( array($identity, 'id', 'user_id') );
       }
       return false;
}

```

sample, if user is` logged_in` with `remember me`
so reaching this with `expire` `session`, it just will return `FALSE` 
then `relogin `again, 

my point is, when `admin` try to set inactive a logged_in user,
the logged in user will still logging in success if `remember`

the `recheck_session` will not fully function with `remember `logged_id

hmmm Im am right with this? or any suggest

by the way thanks for time for understanding @avenirer , and @benedmunds .

avenirer:
@lloricode, as I said in the comment, you did change behaviour by moving that from the constructor (before the PR, the loading of the library would have also logged in the users that checked the remember me check box). But in my opinion you changed it in a good way (even if not where you wanted it to be changed) but what you wrote just now can be part of another issue (a big issue, if you ask me... so I will also look into that), not related to this particular PR.

lloricode:
@avenirer ok sir, appreciated  :), thanks by the way.

PR #1079, opened 2017-05-03
Let's make sure the user it's active

avenirer:
If we want to login a remembered user, we should also make sure that the user is active. This should address the problem reported by @lloricode in PR #1078.


benedmunds:
This looks good, just need one change.  Change it from '1' to 1 so it's passing an integer.

avenirer:
@benedmunds done

benedmunds:
Merged, thanks!

avenirer:
thank you.

PR #1092, opened 2017-06-12
 Use hash_equals() for hash verification

### LINKED ISSUE #1089: Security issue : use hash_equals() instead of '===' to compare password hashes
Dear dev,

First of all, thanks for this library. It is really good.

I would like to point out a security issue in the BCrypt class : 
```php
  public function verify($input, $existingHash) {
    $hash = crypt($input, $existingHash);
    return $hash === $existingHash;
  }
```
A simple strict equals sign `===` is used for hash comparison, which is vulnerable to timing attack.

The `hash_equals()` function should be used (http://php.net/manual/en/function.hash-equals.php) for comparing hashes (as recommended in http://php.net/manual/en/function.crypt.php).

This function is available for PHP >= 5.6. For older PHP version, there are a lot of functions in the comments of the `hash_equals()` function.

Regards

Indigo744:
Fix #1089

The `hash_equals()` function should be used (http://php.net/manual/en/function.hash-equals.php) for comparing hashes (as recommended in http://php.net/manual/en/function.crypt.php).

In this implementation, I've added a private method `hashEquals()`.
If a `hash_equals()` exists, it will be used (case of PHP 5.6 or newer, or CI3).
If no such function exists, (case of CI2 with PHP 5.5 or older), the verification is made using the same algorithm as `hash_equals()` compat function from CI3.

### DIFF
--- a/libraries/Bcrypt.php
+++ b/libraries/Bcrypt.php
@@ -38,7 +38,7 @@ class Bcrypt {
      */
   public function verify($input, $existingHash) {
     $hash = crypt($input, $existingHash);
-    return $this->_hash_equals($existingHash, $hash);
+    return $this->hashEquals($existingHash, $hash);
   }
   
    /**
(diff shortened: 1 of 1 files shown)


benedmunds:
Thanks so much!

Issue #1091, opened 2017-06-12
[Postgres] Wrong users_groups insert when registering user


phiwui:
**_Context :_**
I migrate and old application (pure uggly PHP and mysql DB) into CI with Ion_Auth on Postgre DB.

In order to migrate the users of the old app, i use a migration script which get the user info from the old database and call the register method.

**_Bug description :_**
The problem comes from the use of the method insert_id (line 935 of Ion_auth_model.php) that returns a wrong id and use it to insert into the users_groups table.

According to CI documentation (https://www.codeigniter.com/user_guide/database/helpers.html):
If using the PDO driver with PostgreSQL, or using the Interbase driver, this function requires a $name parameter, which specifies the appropriate sequence to check for the insert id.

**_Proposed solution :_**
By modifying this line with :
$id = $this->db->insert_id($this->tables['users'] . '_id_seq');

the insert into users_groups table works as expected.

There is other lines using insert_id method in this file, there might also need to be modified.

_Sorry for my poor english but it's not my native language._

benedmunds:
Would it be possible to just check for this case in the code and make this conditional?  To support both Postgres+PDO and others?

phiwui:
Acording to the code of this method in others drivers, i think that passing this argument is not a problem. Other db drivers simply ignore it.

benedmunds:
Ah nice!  Send a PR then and I'll be happy to merge.

phiwui:
The default branch seems to be named '2'.

From which branch may I must base my PR 'master' or '2' ?

Edit: I assum that is the brach '2'. Tell me if it's wrong.

benedmunds:
Closing since this was merged in.  Thanks!

PR #1093, opened 2017-06-13
Fix Bug #1091 : Add the name of the sequence for insert_id method

phiwui:
I only set the name of the sequence for Postgres database.

For IBase (which I nerver use), I don't know the default name of the sequence, It is possible to use a if statement with $this->db->dbdriver (pdo or ibase) and/or  $this->db->subdriver (pgsql) to determine the database and set the argument with the good value. But, I twonder if some people really use it with Ion_Auth.

For other db, this PR doesn't change anything (I test it with pdo_mysql) : the argument passed to insert_id is ignored.

### DIFF
--- a/models/Ion_auth_model.php
+++ b/models/Ion_auth_model.php
@@ -932,7 +932,7 @@ class Ion_auth_model extends CI_Model
 
 		$this->db->insert($this->tables['users'], $user_data);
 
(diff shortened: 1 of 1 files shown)


benedmunds:
Looks great, thank you!

Issue #1094, opened 2017-06-15
Timestamp should be taken into account when getting attempts number


Indigo744:
Dear dev,

The `get_attempts_num()` function of Ion_auth_model retrieve all attempts for a specific identity.
I expected this function to actually retrieve the number of attempts from the last `lockout_time` seconds.

It makes sense, my clients actually expect that if they wait enough time (i.e. the `lockout_time` config value), they will have all their "try" back (i.e. the `maximum_login_attempts` config value).

So may I suggest to only retrieve attempts made in the last `lockout_time` seconds?
I understand this behavior is maybe not for everyone, so maybe this can be configurable?

What are your thoughts on this?

Thanks.
Regards.

Indigo744:
After some digging, I saw this was (partially) discussed in #859.

benedmunds:
If you want to send a PR to make that configurable I'll consider it but not sure how I feel about that.  The point of this feature is to prevent brute forcing and more you do to slow down logins the better for that.  I would think that after X tries a user really should just do the forgot your password workflow rather than attempting a dozen passwords.

You could always do a hook to call clear_login_attempts for your specific app implementation.

Indigo744:
I see what you mean.

However, speaking of `clear_login_attempts()` function, it has an interesting behavior.
It will clear all login attempts related to a specific identity (and IP address is option is set), **AND** it will clear _all other expired login attempts_.

This means, if I use `clear_login_attempts()` for user A (either using a hook, or naturally because he successfully logged in), and user B was locked-out and was trying one login at a time, user B will be able to suddenly try multiple times again.
I'm not sure I'm explaining it well, but do you see what I mean?

benedmunds:
Yea I see what you mean there.  Reviewing this PR I'm tending to agree with your thoughts here.  It's a lot simpler and more consistent.  I'll merge it in.  Thanks!

benedmunds:
Merged #1095

Thanks again!

Indigo744:
Waaaaait the PR wasn't actually directly related to this issue XD
The PR is only for having a consistent check behavior.

Here, I'm just talking about whether the `get_last_attempt_time()` function should also check the timestamp field...

benedmunds:
My bad!  I saw you had updated get_last_attempt_time.  Reopening :D

Indigo744:
[...]

PR #1096, opened 2017-06-15
Use timestamp when retrieving the number of attempts

Indigo744:
As discussed in #1094.

The principal modification concerns the `get_attempts_num()` function, which now filter by timestamp value as well.
The nice thing is that `is_time_locked_out()` function is now redundant since the number of attempts will already take into accounts the lockout_time.
Hence, we could actually remove `is_time_locked_out()` as well as `get_last_attempt_time()` (which was made only for this). However, we keep them for retrocompatibility purpose, as some user may rely on them. 

The `clear_login_attempts()` function is updated as well to make sure the expire period value, used to remove older attempts, is not set smaller than the lockout_time configuration value as this would be a security flaw.

### DIFF
--- a/config/ion_auth.php
+++ b/config/ion_auth.php
@@ -94,10 +94,11 @@ $config['manual_activation']          = FALSE;               // Manual Activatio
 $config['remember_users']             = TRUE;                // Allow users to be remembered and enable auto-login
 $config['user_expire']                = 86500;               // How long to remember the user (seconds). Set to zero for no expiration
 $config['user_extend_on_login']       = FALSE;               // Extend the users cookies every time they auto-login
-$config['track_login_attempts']       = FALSE;               // Track the number of failed login attempts for each user or ip.
+$config['track_login_attempts']       = TRUE;                // Track the number of failed login attempts for each user or ip.
 $config['track_login_ip_address']     = TRUE;                // Track login attempts by IP Address, if FALSE will track based on identity. (Default: TRUE)
 $config['maximum_login_attempts']     = 3;                   // The maximum number of failed login attempts.
-$config['lockout_time']               = 600;                 // The number of seconds to lockout an account due to exceeded attempts
+$config['lockout_time']               = 600;                 /* The number of seconds to lockout an account due to exceeded attempts
+							        You should not use a value below 60 (1 minute) */
 $config['forgot_password_expiration'] = 0;                   // The number of milliseconds after which a forgot password request will expire. If set to 0, forgot password requests will not expire.
 $config['recheck_timer']              = 0;                   /* The number of seconds after which the session is checked again against database to see if the user still exists and is active.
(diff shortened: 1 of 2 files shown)


benedmunds:
This looks good.  Want to add the clear_login_attempts change here before I merge or want to do a separate PR?

Indigo744:
Actually, I've only updated the `clear_login_attempts()` function to disallow expire value lower than lockout_time (which would be a security issue, as attempts would be purged while locking still active).

I think the default value of 24h makes sense. It keeps the former behavior, and actually could be helpful to troubleshoot login problem of a client in the same day.

So now it's more secure in case the dev set the lockout_time to higher than 24h and didn't change the parameter.

Is it ok ?

Indigo744:
Also, as I said, `is_time_locked_out()` and `get_last_attempt_time()` are no longer really needed. Do you agree we need to keep them for compatibility purpose?

Especially about `is_time_locked_out()`, which is now only a wrapper for `is_max_login_attempts_exceeded() `...

benedmunds:
That works for me.  Thanks for this, it definitely makes the code simpler and the process easier for users to understand.

benedmunds:
Yes I would like to keep them. Just because the library isn't using it doesnt mean users might not be and I try my best not to break BC

Indigo744:
That's what I thought :)

config/ion_auth.php
```diff
@@ -94,10 +94,11 @@
 $config['remember_users']             = TRUE;                // Allow users to be remembered and enable auto-login
 $config['user_expire']                = 86500;               // How long to remember the user (seconds). Set to zero for no expiration
 $config['user_extend_on_login']       = FALSE;               // Extend the users cookies every time they auto-login
-$config['track_login_attempts']       = FALSE;               // Track the number of failed login attempts for each user or ip.
+$config['track_login_attempts']       = TRUE;                // Track the number of failed login attempts for each user or ip.
 $config['track_login_ip_address']     = TRUE;                // Track login attempts by IP Address, if FALSE will track based on identity. (Default: TRUE)
 $config['maximum_login_attempts']     = 3;                   // The maximum number of failed login attempts.
-$config['lockout_time']               = 600;                 // The number of seconds to lockout an account due to exceeded attempts
+$config['lockout_time']               = 600;                 /* The number of seconds to lockout an account due to exceeded attempts
```
lloricode, line 100: @benedmunds @Indigo744 , maybe need to validate the time set up,

if below 60 sec, just warning/show_error on dev envi that this will need to change.

is this ok?

benedmunds: I'd rather just warn the developer in the comments and let them set what they prefer unless I'm missing something here?  I don't like forcing things on users unless I have to.

lloricode: @benedmunds you have a point :)

PR #1097, opened 2017-06-15
Only use IP if track_login_ip_address is set

Indigo744:
The functions `increase_login_attempts()` and `clear_login_attempts()` made use of the IP address, independently of the `track_login_ip_address` configuration value.

So we make sure that the setting is correctly used.

### DIFF
--- a/models/Ion_auth_model.php
+++ b/models/Ion_auth_model.php
@@ -1172,8 +1172,11 @@ class Ion_auth_model extends CI_Model
 	 **/
 	public function increase_login_attempts($identity) {
 		if ($this->config->item('track_login_attempts', 'ion_auth')) {
-			$ip_address = $this->_prepare_ip($this->input->ip_address());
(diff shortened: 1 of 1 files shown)


benedmunds:
I think we'd want to store the IP in the attempts table either way.  Then just not use it if it's not setup for it.

It doesnt hurt to track it and it comes in handy for troubleshooting in my experience.

Thoughts?

Indigo744:
I would tend to agree. However, there are some laws (like the European data protection law) which forbid to store any user information if he hasn't explicitly agreed to.

I know this is far-fetched in some case, but if the dev don't want the IP to be tracked at login attempts, why should we do it anyway?

This is also to be consistent with the other functions...

benedmunds:
Yea I can see your point there.  Who knows how those laws will change in the future either 😱

benedmunds:
Merging

avenirer:
Security centric cookies are clearly exempt from consent according to the EU advisory body on data protection- WP29. (http://ec.europa.eu/ipg/basics/legal/cookies/index_en.htm)

Issue #1098, opened 2017-06-15
Get Login attempts information for other users


Indigo744:
Hello,

First of all, this is _not_ related to the current PR (just to be sure ^^).

So, let's me explain some context.
In an application I am currently developing, the administrators have access to all users information in the backend.
Among these information, I want the administrators to be able to see if a user has made some unsuccessful login attempts: how many, if they are locked, and the ability to remove this lock.

Now all can be done using the following functions:
 - `get_attempts_num($identity)`
 - `is_max_login_attempts_exceeded($identity)`
 - `get_last_attempt_time($identity)`
 - `clear_login_attempts($identity)`

As you can see, all these functions accept the identity as a parameter. So I should be able to use it for a specific user.
However there is an issue. If the `track_login_ip_address` configuration value is set to TRUE (as in my configuration), all these functions will then use the _current_ IP Address as a filter.
This is hardcoded. So I sadly cannot use these function as is.

I have thought of several approach:
 1. _No update of library_: make wrapper functions to set `track_login_ip_address` to FALSE before calling these functions, and then set it back to TRUE.
 2. _No update of library_: rewrite these functions as needed in another app model.
 3. _Update the library_: add a public `$ip_address` attribute to the class, initialized to the current IP. 
 4. _Update the library_: add a new `$ip_address` parameter for each of these functions. Default behavior would be to use the current IP to keep compatibility.

I have a preference for the last approach, which seems to be the cleanest (and may benefit others who may want to do the same).

What do you think?

benedmunds:
I prefer option 4 and would be happy to merge that in.  I like passing arguments in when possible.

Indigo744:
I guess I can close this now :) thanks.

benedmunds:
Thanks again!!!

PR #1099, opened 2017-06-16
Add IP parameter for login attempts functions

Indigo744:
As discussed in #1098.

The following functions are updated:
 - `get_attempts_num()`
 - `is_max_login_attempts_exceeded()`
 - `is_time_locked_out()`
 - `get_last_attempt_time()`
 - `clear_login_attempts()`

These functions now will accepts an additional and optional parameters, `$ip_address`.
This parameter is obviously only used when `track_login_ip_address` is TRUE.
If NULL (_default value_), the current IP address is used. It matches the former behavior.
If not NULL, the value will be used.

A new function is also added:
 - `get_last_attempt_ip($identity)`
This function will retrieve the last IP address used for a given identity.

### DIFF
--- a/models/Ion_auth_model.php
+++ b/models/Ion_auth_model.php
@@ -1083,14 +1083,18 @@ class Ion_auth_model extends CI_Model
 	 * is_max_login_attempts_exceeded
 	 * Based on code from Tank Auth, by Ilya Konyukhov (https://github.com/ilkon/Tank-Auth)
 	 *
-	 * @param string $identity
+	 * @param string $identity: user's identity
+	 * @param string $ip_address: IP address
+	 *                            Only used if track_login_ip_address set to TRUE.
+	 *                            If NULL (default value), current IP address is used.
+	 *                            Use get_last_attempt_ip($identity) to retrieve user's last IP
 	 * @return boolean
 	 **/
-	public function is_max_login_attempts_exceeded($identity) {
+	public function is_max_login_attempts_exceeded($identity, $ip_address = NULL) {
 		if ($this->config->item('track_login_attempts', 'ion_auth')) {
 			$max_attempts = $this->config->item('maximum_login_attempts', 'ion_auth');
 			if ($max_attempts > 0) {
-				$attempts = $this->get_attempts_num($identity);
+				$attempts = $this->get_attempts_num($identity, $ip_address);
 				return $attempts >= $max_attempts;
 			}
 		}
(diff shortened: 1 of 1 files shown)


models/Ion_auth_model.php
```diff
@@ -1141,12 +1153,16 @@ public function is_time_locked_out($identity) {
...
+	 *                            Only used if track_login_ip_address set to TRUE.
+	 *                            If NULL (default value), current IP address is used.
+	 *                            Use get_last_attempt_ip($identity) to retrieve user's last IP
+	 * @return int
 	 */
-	public function get_last_attempt_time($identity) {
+	public function get_last_attempt_time($identity, $ip_address = NULL) {
 		if ($this->config->item('track_login_attempts', 'ion_auth')) {
-			$ip_address = $this->_prepare_ip($this->input->ip_address());
+			if (is_null($ip_address)) $ip_address = $this->_prepare_ip($this->input->ip_address());
```
benedmunds: Can you re-format this to match the rest of the styling.  Also use isset instead of is_null in case the user sends a falsey value.  For formatting, either with curlys or as a ternary:

    if (!isset($ip_address)) 
    {
        $ip_address = $this->_prepare_ip($this->input->ip_address());
    }


or

    $ip_address =  isset($ip_address) ? $ip_address : $this->_prepare_ip($this->input->ip_address());

models/Ion_auth_model.php
```diff
@@ -1185,19 +1224,23 @@ public function increase_login_attempts($identity) {
...
-	public function clear_login_attempts($identity, $old_attempts_expire_period = 86400) {
+	public function clear_login_attempts($identity, $old_attempts_expire_period = 86400, $ip_address = NULL) {
 		if ($this->config->item('track_login_attempts', 'ion_auth')) {
 			// Make sure $old_attempts_expire_period is at least equals to lockout_time
 			$old_attempts_expire_period = max($old_attempts_expire_period, $this->config->item('lockout_time', 'ion_auth'));
 
 			$this->db->where('login', $identity);
 			if ($this->config->item('track_login_ip_address', 'ion_auth')) {
-				$ip_address = $this->_prepare_ip($this->input->ip_address());
+				if (is_null($ip_address)) $ip_address = $this->_prepare_ip($this->input->ip_address());
```
benedmunds: Same as above comment ^

benedmunds (changes requested):
This looks great, just have a couple formatting preferences to try to keep the code uniformly formatted (although I don't always do the best job of that).

models/Ion_auth_model.php
Indigo744: Agreed.

models/Ion_auth_model.php
Indigo744: Agreed.

benedmunds (approved):

benedmunds:
Awesome, thanks!

Indigo744:
Glad to help :)

PR #1132, opened 2017-09-02
Stop truncating IPv6 addresses, bumped max email char limit & other migration tweaks

DakuTree:
Few useful tweaks here:
* ip_address columns now have a constraint of 45, as to not truncate IPv6 addresses. (As per: https://stackoverflow.com/a/166157/1168377)
* email constraint has been bumped from 100 to 254, which is the max possible length of an email address (As per: https://stackoverflow.com/a/574698/1168377)
* Various migration tweaks: Config is now used for table names, salt can be null (which is what it already is in the .sql files), make sure dbforge is loaded (as not everybody autoloads it) & some formatting changes for readability.


benedmunds:
These changes are great, thanks so much!

PR #1148, opened 2017-11-02
Use (only) ->limit(1) in the check functions

aphofstede:
No need for grouping or ordering here, since we just need >0.


benedmunds:
Thanks!

PR #1149, opened 2017-11-03
Improved white-space, formatting and PHPDoc/comments

aphofstede:
Changed to be more consistent with either the rest of the lib or CI itself.
Note: Added an @deprecated tag to `_prepare_ip()`, `is_time_locked_out()` and `get_last_attempt_time()`, since the comments already said as much.

### DIFF
--- a/config/ion_auth.php
+++ b/config/ion_auth.php
@@ -1,25 +1,25 @@
-<?php  if ( ! defined('BASEPATH')) exit('No direct script access allowed');
+<?php
 /**
-* Name:  Ion Auth
-*
-* Version: 2.5.2
-*
-* Author: Ben Edmunds
-*		  ben.edmunds@gmail.com
-*         @benedmunds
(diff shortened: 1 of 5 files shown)


benedmunds:
Really great update here, thanks for this.

One item, the doc blocks for the user_ids should have float as a possible type as well shouldnt they?

aphofstede:
Yes, I guess you're right. Although it's a bit silly, if they get cast to float anyway you can also pass them in. I'll update the PR.

aphofstede:
Done. Note that group/user ID is not cast to float everywhere; was that intentional? I don't use bigint myself.

benedmunds:
Shouldn’t really matter as long as it’s casted before it gets to the DB, and even then the DB will catch it but I try to avoid that.

Thanks!

benedmunds:
Merged 🎉

PR #1153, opened 2017-11-09
mapping out directory / file structure for setup

gegere:
How does this look? :)


benedmunds:
This looks great!  Can you leave the paragraph about being about to just copy to corresponding folders though, or maybe do an example of that?  I just don't want people to think third party is the only way it'll work.  What do you think is best?

gegere:
Ready for your review!

benedmunds:
Dude this is great, thank you!

PR #1158, opened 2017-11-16
Update Ion_auth_model.php

CharbelJ:
You should not be able to delete yourself if your are the current logged in user.

### DIFF
--- a/models/Ion_auth_model.php
+++ b/models/Ion_auth_model.php
@@ -1913,20 +1913,29 @@ class Ion_auth_model extends CI_Model
 
 		$this->db->trans_begin();
 
-		// remove user from groups
-		$this->remove_from_group(NULL, $id);
-
-		// delete user from users table should be placed after remove from group
-		$this->db->delete($this->tables['users'], array('id' => $id));
-
-		if ($this->db->trans_status() === FALSE)
+		// do not allow the current logged in user to delete him/her self
+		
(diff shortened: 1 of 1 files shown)


benedmunds:
Hey,

For for this but currently there's code in the example controller to prevent this.  I don't think it's something we want to stop at the model level because there are valid use cases where a user should be able to delete themselves.  

For example, in the case of "delete my account and all my data" the user should be able to delete their account and then be logged out afterwards.  Which is really something that should be handled at the controller level.

CharbelJ:
So I should be checking if a user ir currently logged in and deleting themselves on the controller level yeah?

benedmunds:
Hmm I thought there was an example of this in the example controller but there's not..

Yea you'd want to do this in your controller.

CharbelJ:
Thanks a lot man!

CharbelJ:
Would you please show me an example how to handle it in the controller?

benedmunds:
I imagine you have a delete method in your controller or something similar..  If so it'd be something like this:

`
function delete ($id) {
    if ($this->ion_auth->user()->row()->id == $id) {
       show_error('You can not delete yourself');
    }

  .//rest of your logic
`

You'll want to update it to use your validation or whatever you're doing to return errors.

CharbelJ:
Perfect just wanted to see if I was on the right track and it turned I was! Thanks a lot man! Really appreciate it!

benedmunds:
No problem, take care!

PR #1159, opened 2017-11-16
Update ion_auth_lang.php

CharbelJ:


benedmunds:
See comments in #1158

PR #1165, opened 2017-11-28
Update for Ion auth Model

Premgupta01:
just added code  for check Old and New Password should not be same

### DIFF
--- a/models/Ion_auth_model.php
+++ b/models/Ion_auth_model.php
@@ -1,171 +1,164 @@
-<?php
+<?php  if ( ! defined('BASEPATH')) exit('No direct script access allowed');
 /**
- * Name:    Ion Auth Model
- * Author:  Ben Edmunds
- *           ben.edmunds@gmail.com
- * @benedmunds
- *
- * Added Awesomeness: Phil Sturgeon
- *
- * Created:  10.01.2009
- *
- * Description:  Modified auth system based on redux_auth with extensive customization. This is basically what Redux Auth 2 should be.
- * Original Author name has been kept but that does not mean that the method has not been modified.
- *
- * Requirements: PHP5 or above
- *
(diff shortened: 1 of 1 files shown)


benedmunds:
Couple bits of feedback.

1. Can you remove the whitespace updates that arent related to this change

2. Can you remove the comment update are arent related to this change

3. Make the new matches old error be optional based on config for backwards compatibility.  Default it to false so existing installs don't change.

lloricode:
@Premgupta01 its too difficult to review where is the changes

Premgupta01:
I have written a comment in ion model file and comment is "// just added code given below for check Old and New Password should not be same"

You can easily find out this

Premgupta01:
@benedmunds 

Thanks for reviewing code, I am naive in Codeigniter, So can you please guide me here 

1. Make the new matches old error be optional based on config for backwards compatibility. Default it to false so existing installs don't change.

Do I need to add this change to the latest code?

lloricode:
maybe, just push some related to this PR, to easily check by the owner. its good you format the codes, but, it may cause some trouble if there accidentally change without knowing.

Premgupta01:
okay, @lloricode I am naive and trying to improve my asset with the help of you and others,
Thanks for letting me know So can you please guide me How to add new code to anyone's existing Code.

lloricode:
@benedmunds said that you can or need add a additional config, to set if where need to check if password is same on old. then make it optional so the other using with this library will not break there code if ever merge this PR

Premgupta01:
Yes, I have added it to the config file for display Error Message.

benedmunds:
Closing this due to age.  @Premgupta01 if you want to submit a new PR with *only* the one change you are suggesting and without all the comment and whitespace changes I'll be happy to consider it.  Thanks!

PR #1168, opened 2017-12-01
Remove id fields from insert data sets

qury:
Remove id fields from insert data sets as it is causing issues with PostgreSQL's internal sequence.


benedmunds:
Awesome, thank you!

PR #1184, opened 2018-02-09
Update: Possibility of changing the column names easily

LuanBaviloni:
Implementation of the possibility of changing the column names of tables.

### DIFF
--- a/config/ion_auth.php
+++ b/config/ion_auth.php
@@ -42,6 +42,46 @@ $config['join']['users']  = 'user_id';
 $config['join']['groups'] = 'group_id';
 
 /*
+| -------------------------------------------------------------------------
+| Fields of tables.
+| -------------------------------------------------------------------------
+| Database field names.
+*/
+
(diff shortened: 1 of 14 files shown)


benedmunds:
Can some people thumbs up or down this based on if they would use this functionality?  

I’ve only seen one or two people ask for this over the years that I remember...

ghosthouse:
Yes would like that option very much.

avenirer:
Then, maybe, the next thing would be to also give the possibility to change the name of the tables... I guess this is getting out of hand. :)

LuanBaviloni:
It's useful when you have to follow specific naming patterns/conventions in database modeling.

paklie:
thumbs up..

benedmunds:
Closing this PR since we didn't get overwhelming votes for it and it's not worth the added complexity imho.  If you want to submit a similar PR against the v4 branch, or just commenting and getting @bvrignaud 's thoughts in #1273 I would consider implementing it there since it'l be a fresh release.

PR #1194, opened 2018-03-18
Feature: allow to specify a DB group name

Indigo744:
For one of my project, I have several database connections defined in the `database.php` config file.
The one to use depends on several factors, but it is not unusual to have two or more connections loaded in CI (e.g. the first one is accessible by `$this->db`, and the second one by `$this->db_special`, and so on...).

Unfortunately, IonAuth may not be on the default connection. And I don't have any way to configure it.

Hence this PR. I propose a new option `database_group_name` in which the user can set the DB connection group.
This feature is fully backward compatible. It will work as before in its default configuration.

Under the hood, the `load_database()` call is a bit more complex and directly assign it to a local `$this->db` member (instead of using the one loaded globally). But the nice thing is that no change is needed for existing DB call.

**As an additional benefit**, this PR fix a potential bug in which the user has deactivated the Query Builder option globally (`$query_builder` option in `database.php`). Since IonAuth use it extensively, it would break. However, this feature call the database loader with the `$query_builder` option overrided.

### DIFF
--- a/config/ion_auth.php
+++ b/config/ion_auth.php
@@ -22,6 +22,17 @@
 defined('BASEPATH') OR exit('No direct script access allowed');
 
 /*
+ | -------------------------------------------------------------------------
+ | Database group name option.
+ | -------------------------------------------------------------------------
+ | Allows to select a specific group for the database connection
+ |
+ | Default is empty: uses default group defined in CI's configuration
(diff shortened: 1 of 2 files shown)


benedmunds:
This is a great addition, thanks so much! Merging.

Issue #1195, opened 2018-03-19
bcrypt / hash function evolution


Indigo744:
Dear @benedmunds 

First and foremost, thank you again for this lib. I use it extensively and I love it  :smile: This is why I try to give back by proposing updates and enhancement when I find the time.

The usage of the specific bcrypt library has always bugged me. Of course I've audited the code and found out that it was OK using `crypt()` and such, but I keep thinking that it could be better.

I found out that I was not the only one (see related issues #1176 and #1169). 

Another thing is using `crypt()` function will not allow us to provide an option to use the new ARGON2 algorithm, which is available in PHP 7.2 but only for the `password_hash()` function.

Hence I would like to make a PR to move to using `password_*` functions.

But before starting to work on it, I have a couple of questions I would like to pass by you first:
1. Should PHP < 5.5 still be supported? [According to Wikipedia](https://en.wikipedia.org/wiki/PHP#Release_history), PHP 5.3 is not supported since 2014, and PHP 5.5 since 2016... 
Should an Auth library really support this old version?
2. If PHP < 5.5 is needed, should we use [password_compat](https://github.com/ircmaxell/password_compat)? 
How should we load it? Copy/pasting the code would seem easy, but it would be troublesome to maintain (checking for updates regularly, etc...). 
Also, CI 3 offers their own polyfills (see `core/compat/password.php`) so it would be quite redundant...
So maybe we should simply redirect users to this lib and recommends them to install it? I really wonder the number of users with CI 2 and PHP < 5.5...

Thanks for your insights.

Indigo744:
Also, there are two differents salt generator functions implemented: `Ion_auth_model::salt()` and `Bcrypt::getSalt()`...

benedmunds:
Hey @Indigo744 

Pretty excited that you're up for taking this on.  

I'm thinking that this would be worth cutting a new major version of Ion Auth (v3) and then you could just support CI3 and rely on the polyfills being present.  

This versioning makes sense and will support adding a CI 4 version later as well.

What are your thoughts on the above?

Indigo744:
So it would mean we should target CI 3 and PHP 5.5?
[CI 3 server requirement](https://www.codeigniter.com/user_guide/general/requirements.html) ask for PHP 5.6, but suggest 5.3.7 would work (but strongly advised against).

I think that would be a good thing indeed, and would allow to clean up some old legacy code.

[...]

PR #1196, opened 2018-03-19
IonAuth 3: Drop internal Bcrypt and use password_* PHP function

Indigo744:
As said in #1195 

Drop internal Bcrypt and use password_* PHP function.
The `salt_prefix` option is drop as no longer needed.

Also, default rounds was increased for better security by default (computer are more powerful now than it was when IonAuth was created 😸)


models/Ion_auth_model.php
```diff
@@ -268,15 +259,26 @@ public function __construct()
...
 			return FALSE;
 		}
 
 		// bcrypt
-		if ($use_sha1_override === FALSE && $this->hash_method == 'bcrypt')
+		if ($use_sha1_override === FALSE)
 		{
-			return $this->bcrypt->hash($password);
+			// Get cost
+			if ($this->random_rounds)
```
benedmunds: Maybe we should just remove the random rounds?  After a certain point it's really not adding much to the security when using bcrypt...

models/Ion_auth_model.php
```diff
@@ -325,14 +327,9 @@ public function hash_password_db($id, $password, $use_sha1_override = FALSE)
 		}
 
 		// bcrypt
-		if ($use_sha1_override === FALSE && $this->hash_method == 'bcrypt')
+		if ($use_sha1_override === FALSE)
```
benedmunds, line 324: We can remove the sha1 override completely right?

models/Ion_auth_model.php
```diff
@@ -268,15 +259,26 @@ public function __construct()
 	 */
 	public function hash_password($password, $salt = FALSE, $use_sha1_override = FALSE)
 	{
-		if (empty($password))
+		// Check for empty password, or password containing null char
+		if (empty($password) || strpos($password, "\0") !== FALSE)
```
benedmunds: maybe just do a !isset() here?  I'm not sure what all the null char check might be for here.

models/Ion_auth_model.php
```diff
@@ -182,6 +182,14 @@ public function __construct()
 		$this->load->helper('date');
 		$this->lang->load('ion_auth');
 
+		// PHP password_* function sanity check
+		if (!function_exists('password_hash') || !function_exists('password_verify'))
+		{
+			show_error("PHP function password_hash or password_verify not found. " .
+						"Are you using CI 2 and PHP < 5.5? " .
+						"Please upgrade to CI 3, or PHP >= 5.5, or use password_compat.");
```
benedmunds: Throw a link to password_compat in this message please - or reference a link thats in the readme and add it there.  I'm sure a lot of people won't know what it is offhand.

models/Ion_auth_model.php
Indigo744: Agreed, let's simplify it 😃

models/Ion_auth_model.php
Indigo744: I'll make another PR regarding sha1 removal. I didn't want to modify everything in one PR ;-)

models/Ion_auth_model.php
Indigo744: Sorry, I should have explained a bit more.
See this comment on doc: http://php.net/manual/en/function.password-hash.php#118603

models/Ion_auth_model.php
Indigo744: Yeah, I spent a lot of time wondering about it actually. I was hoping people would just google it. But you're right. I'll add the link to password_compat.

models/Ion_auth_model.php
benedmunds: Ah I didnt know that, thanks for the info!

models/Ion_auth_model.php
Indigo744: Well, actually I was wrong: this very line already remove the sha1 support!
I've updated the config file accordingly.

The PR I was talking about is more related to the `$use_sha1_override`, and by extension the way salt and token are generated. This will indeed need a dedicated PR as changes are bigger.

PR #1198, opened 2018-03-20
IonAuth 3: Drop CI2 support

Indigo744:
As said in #1195

Drop CodeIgniter 2 support.


models/Ion_auth_model.php
```diff
@@ -182,6 +182,11 @@ public function __construct()
 		$this->load->helper('date');
 		$this->lang->load('ion_auth');
 
+		// Sanity check for CI2
+		if (substr(CI_VERSION, 0, 1) === '2') {
```
benedmunds: Maybe move both this and the password_compat checks to a new method checkCompatibility() or something like that, and call it from the library constructor.

Indigo744: I was indeed wondering where to put these checks. You prefer the library? I understand :)

benedmunds:
👏👏👏

Issue #872, opened 2015-10-29
forgotten password complete email


iamthestreets:
Hi,

I noticed there is a function in the `Ion_auth` library called `forgotten_password_complete`, but it does not seem to be used. Also, if it were to be used it would not work unless I'm missing something. The function requires the forgot password code to passed `public function forgotten_password_complete($code)` , but the code is removed once the password has been reset.

So I guess my first question is does the `forgotten_password_complete` function get used?
If no would i need to completely rewrite it since the forgot password is being removed before I can call it?

avenirer:
Indeed it seems that that function is no longer used. Maybe it was used in the passed and now it's kept for compatibility reasons @benedmunds ? Shouldn't we delete it? But what do you mean @iamthestreets  by "I need to completely rewrite it"? Do you need to do something special after someone resets the password?

iamthestreets:
I want to send an email letting the user know their password has been changed. There could be many reasons why you should do this, the most important is to make sure they are aware of the change just incase it wasn't them. 

What I mean by rewriting is the current `forgotten_password_complete` function requires you to pass the forgotten password code to the function to trigger the email to be sent, but by the time the change is complete to code no longer exists. Also, it is designed to send the actual password in an email which should not be done for security reasons. 

If I have to rewrite it myself I am ok with that I just want to make sure I'm not missing something and I would also suggest to delete it if it is no longer used or rewrite it.

Thanks!

benedmunds:
Yea it's not used anymore but older versions did so kept it for BC.

@iamthestreets you'd probably be best using a hook for this: http://codebyjeff.com/blog/2013/01/using-hooks-with-ion_auth

iamthestreets:
I somewhat understand the hooks and trigger events, but i still don't see the benefit of using it. I also don't know how it would work for me in this case. When/where do I set the hook? In the Auth `reset_password()` function? Also, as I mentioned before the code needs to be passed to the `forgotten_password_complete()` function so correct me if i'm wrong, but this function will not work if I want to wait for the reset to be completed because the code will be removed.

Can you help with an example?

benedmunds:
You'd be best just writing a completely new function for sending that confirmation email, then call it after activation (via a hook) to send the email.

PR #1199, opened 2018-03-20
Ion Auth 3 : cleanup unused functions

Indigo744:
Related to #1195

Some functions were kept for BC. They should be removed.

1. Function `forgotten_password_complete()` doesn't seem to be used anymore, and has a security issue since the password is sent by mail in clear text (see #872) 
2. Function `_prepare_ip()`: the doc says it is only kept for BC
3. Function `is_time_locked_out()`: doc says it is only a wrapper for `is_max_login_attempts_excedeed()`

Do you see any other functions to remove?


benedmunds:
Good stuff, I love seeing old code removed.  I don't see anything else that should be removed.  

Can you updated the PR to be agains the 3 branch.

Indigo744:
Done 😺

benedmunds:
Still looks like its against the 2 branch 
screen shot 2018-03-20 at 10 49 38 am

Indigo744:
Oh damn sorry!

benedmunds:
No problem! Thanks

PR #1200, opened 2018-03-20
Ion Auth 3: remove salt and better token generator

Indigo744:
Related to #1195 

Since bcrypt handles the salt internally, there is no need to store it separately.

This PR removes any salt-related code.

Also, the salt generator was used to generate the remember and forgotten token.
We greatly simplify it by using a simple `random_token()` generator based on function known to be random enough. In case a user misses the proper function (I would guess less than 1%), then the [random_compat](https://github.com/paragonie/random_compat) library is proposed.

### DIFF
--- a/models/Ion_auth_model.php
+++ b/models/Ion_auth_model.php
@@ -190,8 +190,6 @@ class Ion_auth_model extends CI_Model
 
 		// initialize data
 		$this->identity_column = $this->config->item('identity', 'ion_auth');
-		$this->store_salt = $this->config->item('store_salt', 'ion_auth');
-		$this->salt_length = $this->config->item('salt_length', 'ion_auth');
 		$this->join = $this->config->item('join', 'ion_auth');
 
 		// initialize hash method options (Bcrypt)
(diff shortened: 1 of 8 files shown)


models/Ion_auth_model.php
```diff
@@ -2499,4 +2338,36 @@ protected function _filter_data($table, $data)
...
+			return bin2hex(mcrypt_create_iv($result_length / 2, MCRYPT_DEV_URANDOM));
+		}
+
+		// Try openssl
+		if (function_exists('openssl_random_pseudo_bytes')) {
+			return bin2hex(openssl_random_pseudo_bytes($result_length / 2));
+		}
+
+		// No luck!
+		show_error("Cannot generate a token random enough. " .
```
benedmunds: What you have here is good, but add these function checks and the error to the library compat check as well.  That way the user sees up front that they arent compatibile instead of it popping up when they test the token functionality.

Indigo744: Got it.

PR #1204, opened 2018-03-20
Ion Auth 3: Implements rehash_password_if_needed

Indigo744:
See #1195 

When user log-in, check if password needs to be rehashed according to the current set of parameters.

Allows smooth transition with different algo, or costs...

Also, since quite a lot of different function update the user's password in the DB, in order to avoid code duplication, a new internal function `_set_password_db()` is introduced.


benedmunds:
Great cleanup 👏

PR #1205, opened 2018-03-20
Ion Auth 3: handle legacy SHA1 password

Indigo744:
As #1195 

First `hash_password_db()` uses the identity column instead of ID (so we can pass around the $identity variable without calling the DB).

Secondly, `hash_password_db()` will detect a SHA1 stored password (i.e. not starting with `$`), and run the specific function `_password_verify_sha1_legacy()`.

This function re-implements how the SHA1 algorithm was being set (with the salting).
If the match if successful, then the password is set to use the newest algorithm (just like `password_needs_rehash` actually).

Hope I didn't miss anything. Some test will be needed of course.

### DIFF
--- a/models/Ion_auth_model.php
+++ b/models/Ion_auth_model.php
@@ -305,7 +305,16 @@ class Ion_auth_model extends CI_Model
 			return FALSE;
 		}
 
-		return password_verify($password, $hash_password_db->password);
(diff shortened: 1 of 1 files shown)


benedmunds:
Great work, I don't see any issues from a code review aspect.  Definitely will need to do some testing later.

PR #1206, opened 2018-03-21
Ion Auth 3 : implements different hashing params for administrators

Indigo744:
See #1195 

The [bcrypt specs](https://www.usenix.org/legacy/publications/library/proceedings/usenix99/provos/provos_html/node6.html#SECTION00051000000000000000) suggests to have different costs for different type of account: 

> At the time of publication, the default cost is 6 for a normal user and 8 for the superuser.

The PR implements a new set of hashing parameters for users in admin group.

Note that the `in_group()` function was moved from the library to the model, so we can use it when retrieving parameters. 

Also, a `get_user_id_from_identity()` is introduced to be able to retrieve a user_id from an identity (and as such use it for group check).


benedmunds:
Good catch

PR #1208, opened 2018-03-21
Ion Auth 3: Better and more secure remember me

Indigo744:
#1195 

I'm quite happy with this one.

I've implemented a more secure and simplified version of "remember me".
It is based on this article: https://paragonie.com/blog/2015/04/secure-authentication-php-with-long-term-persistence

This new behavior introduces a new db column called "remember_selector", which is used to identify the user (and also avoid to store the user email in a cookie).

DB side, the remember code is actually stored as a hash to avoid session stealing when the DB is compromised.

Hash comparison is done with `hash_equals` to avoid timing leaks.

Also, only one cookie is now needed.


Indigo744:
Now I wonder if instead of SHA-384 (or whatever the user selected) we should just have use the same hashing algorithm than the passwords?
It would cost more in term of processing power I think. But it would be simpler in the code?
What do you think?

benedmunds:
Wow nice, I'll start reviewing this in a bit.

It would be nice to use the same algo, and more cost is probably a good thing really.

Indigo744:
Next stage will be to implement the same mechanism for the "reset password" feature.
I think I will make another PR for it when this one is merged.

benedmunds:
I don't even have any critiques here 👏 merging!

PR #1209, opened 2018-03-21
Ion Auth 3:  Implements better and more secure password reset 

Indigo744:
#1195 

"Password reset" feature.

Same approach as PR #1208 related to "remember me" feature

A `forgotten_password_selector` is added to the DB.


libraries/Ion_auth.php
```diff
@@ -139,17 +139,21 @@ public function __get($var)
 	 */
 	public function forgotten_password($identity)
 	{
-		if ($this->ion_auth_model->forgotten_password($identity))
+		// Retrieve user information
+		$user = $this->where($this->ion_auth_model->identity_column, $identity)
```
benedmunds, line 143: While I'm thinking about it, should probably add a default index on the default identify in the SQL files/migration.  Thoughts?

models/Ion_auth_model.php
```diff
@@ -679,44 +680,59 @@ public function forgotten_password($identity)
...
+	/**
+	 * Get a user from a forgotten password key.
+	 *
+	 * @param    string $user_code
+	 *
+	 * @return    bool|object
+	 * @author  Mathew
+	 * @updated Ryan
+	 */
+	public function forgotten_password_get_user($user_code)
```
benedmunds: how about changing this method name to "get_user_by_forgotten_password_code"?

I try to opt for more expressive method names now.

models/Ion_auth_model.php
```diff
@@ -1878,23 +1894,35 @@ public function remember_user($identity)
...
+									  'remember_code' => $token->validator_hashed),
+								array($this->identity_column => $identity));
 
 			if ($this->db->affected_rows() > -1)
 			{
-				$this->_set_remember_cookie_info($selector, $validator);
+				// if the user_expire is set to zero we'll set the expiration two years from now.
+				if($this->config->item('user_expire', 'ion_auth') === 0)
+				{
+					$expire = (60*60*24*365*2);
```
benedmunds, line 1912: Change this to 63072000 and add a comment on what it is so we don't always do this in the code for no real reason other than readability.  Yea I know this was my code from before...

libraries/Ion_auth.php
Indigo744: Yes, as well as for both selector tokens (remember_me and password_reset).
I'll update the migration and user guide in another PR.

models/Ion_auth_model.php
Indigo744: Agreed

models/Ion_auth_model.php
Indigo744: Agreed

libraries/Ion_auth.php
Indigo744: However, in the SQL file, we don't know yet which column is the identity one, right?
Also, what happen if the user change the identity column option?
I think, for the identity column, it should remain as is with the info in the README (and maybe add a comment?) so users can implement optimization when needed.

benedmunds: I think it'd be best to go ahead and optimize for the default identity and comment in the config that they should update their indexing if they change it.  In my experience the majority of users stick with the defaults.

Indigo744: Alright, I'll do that.

Indigo744: But in another PR 😅

benedmunds: Yea sounds good, I'll go ahead and merge this one

Indigo744: See PR #1210

PR #1211, opened 2018-03-22
Ion Auth 3: always clear all codes when state change

Indigo744:
#1195 

`clear_forgotten_password_code()` now uses identity instead of code
Added `clear_remember_code()`

When the state change, we should be careful to clear all codes:
 - When user log in: clear forgotten code and remember me (if remember option not set)
 - When user automatically log in through remember me: clear forgotten code
 - When user log out: clear forgotten code and remember me

### DIFF
--- a/controllers/Auth.php
+++ b/controllers/Auth.php
@@ -319,12 +319,14 @@ class Auth extends CI_Controller
 			}
 			else
 			{
+				$identity = $user->{$this->config->item('identity', 'ion_auth')};
+
 				// do we have a valid request?
 				if ($this->_valid_csrf_nonce() === FALSE || $user->id != $this->input->post('user_id'))
 				{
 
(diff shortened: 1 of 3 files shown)


benedmunds:
I shouldve asked this earlier, but whats the reason for using idenity everywhere instead of ID?

The main problem with this that I could see is I know a lot of people allow users to use multiple idenities and just try a login for both. Then use ID for all the other functions.  For example, they might try username and email.

benedmunds (approved):

Indigo744:
Good question. Actually, it was a value that was readily available in most of the function...

I'll try to come up with a list of function that has evolved to use identity, and we'll see the impact.

benedmunds:
I'm going to go ahead and merge this one in.  Can address identity v id in another PR.

PR #1227, opened 2018-04-02
Feature/  call

pabloleonalcaide:
refactor function __call, 
I've changed the multiple if clauses in __call function with switch-case . Hope it be usefull

### DIFF
--- a/controllers/Auth.php
+++ b/controllers/Auth.php
@@ -550,6 +550,15 @@ class Auth extends CI_Controller
 			$this->_render_page('auth/create_user', $this->data);
 		}
 	}
+	/**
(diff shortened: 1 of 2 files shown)


benedmunds:
Not merging this one for a couple reasons.  

1. This PR includes the previously submitted redirect change.

2. I actually prefer IF statements if there are only a few items to check or if an action is being taken.  I think that's better for readability and flexibility later than switch statements.  What you did here is fine, but that's just a personal preference of mine.

Closing this but thanks for submitting it!

pabloleonalcaide:
I understand, thanks!

Issue #1234, opened 2018-05-01
Windows local machine directory separator


bedrijfsportaal:
Goodmorning,

I'm using ion_auth for the very first time, must say it's awesome with enough features buildin. The only issue i'm currently facing is the directory separator, it's not responding to / but only on \\.

Can easly fix this by adding `DIRECTORY_SEPARATOR` instead of hard-coded, don't know if it's overrided when updated so i can start all over again?

**CodeIgniter Ion Auth default**
`$this->_render_page('auth/index', $this->data);`

**My current fix (works)**
`$this->_render_page('auth' . DIRECTORY_SEPARATOR . 'index', $this->data);`

Should this be updated as by request on this repository or will this be added, don't know if it's by default on server settings or only when activated.

Thanks in Advance!

benedmunds:
Are you on Windows?

bedrijfsportaal:
Correct, working on my local machine with MAMP PRO before pushing this to a web-server.

benedmunds:
Ah that makes sense.  Would you be interested in submitting a PR for these changes?

bedrijfsportaal:
Shure no problem at all, will do this in a few seconds!

PR #1235, opened 2018-05-01
Fix for different directories

bedrijfsportaal:
Fix for the different directories on different machines #1234


benedmunds:
Thanks!  Can you submit the same against the 3 branch?

PR #1236, opened 2018-05-01
Revert 1235 patch 1

bedrijfsportaal:


benedmunds:
Is this correct?

bedrijfsportaal:
I've got no idea what i just did, looks like this will revert the fix or something? Sorry don't use Github that much it's all new for me...

PR #1237, opened 2018-05-01
Fix for different directory separator

bedrijfsportaal:
Different directory separators for Windows or Linux, global constant determine what separator is used.


bedrijfsportaal:
@benedmunds this should be what you've asked for, if not please let me know so I can fix.

benedmunds:
Yea perfect, thanks so much!

PR #1238, opened 2018-05-01
IonAuth 3: add max arbitrary size for password (DOS protection)

Indigo744:
Hello,

I'm back with a new PR :smile:

I came across [this article](https://arstechnica.com/information-technology/2013/09/long-passwords-are-good-but-too-much-length-can-be-bad-for-security/) which state that long password can be used as DOS attack vector. 

It was related to Django, the Python web framework:
>    Unfortunately, this complexity can also be used as an attack vector. Django does not impose any maximum on the length of the plaintext password, meaning that an attacker can simply submit arbitrarily large—and guaranteed-to-fail—passwords, forcing a server running Django to perform the resulting expensive hash computation in an attempt to check the password. A password one megabyte in size, for example, will require roughly one minute of computation to check when using the PBKDF2 hasher.
> This allows for denial-of-service attacks through repeated submission of large passwords, tying up server resources in the expensive computation of the corresponding hashes.

The solution the Django team used was to limit the max password length to an arbitrary number of 4096 bytes. I agree it is long enough, even for people using password managers.

Here is a PR adding this max size to limit DOS vector attack.

Also, I've added a check on every `hash_password()` return value to make sure a FALSE value isn't store in the DB.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@ Ion Auth Changelog
 
 ## xx March 2018 - Ion Auth 3
 
+ - **General:**
+    - No longer work for empty password or password above 4096 bytes (DOS protection)
  - **New server requirements:**
     - Drop CodeIgniter 2 support
     - Drop PHP < 5.6 support
--- a/config/ion_auth.php
+++ b/config/ion_auth.php
@@ -126,6 +126,9 @@ $config['argon2_admin_params']		= [
  | min_password_length:		This minimum is not enforced directly by the library.
  | 							The controller should define a validation rule to enforce it.
  | 							See the Auth controller for an example implementation.
+ |
+ | The library will fail for empty password or password size above 4096 bytes.
+ | This is an arbitrary (long) value to protect against DOS attack.
  */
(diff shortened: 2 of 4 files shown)


benedmunds:
Feel free to let me know if my assumptions here are off.

This shouldn’t be an issue for bcrypt since it truncates long password anyway right?

What if instead of this we truncated after X length no matter the hashing algo?

Indigo744:
This would be a correct assumption for bcrypt only.
But the algo can be Argon2, which doesn't have this kind of limitation.

4096 is quite huge for a password, even using a password manager (the biggest I generated was around 600 chars...).

I'm not confortable with the idea of truncating without telling. I think we are better off returning a proper FALSE and let the dev know something is wrong.

Indigo744:
To quote how the [Django framework handled this](https://www.djangoproject.com/weblog/2013/sep/15/security/): 
> To remedy this, Django's authentication framework will now automatically fail authentication for any password exceeding 4096 bytes.

benedmunds:
Fair points, thanks for all the info!  Merging.

PR #1261, opened 2018-08-16
Reuse default DB connection when no group name

### LINKED ISSUE #1258: Ion_auth_model not reusing the default database connection
I don't know if this is a bug or a feature!
CI : 3.1.9
Ion_auth : 3 & 2

Applies to the following code segment

```
/models/Ion_auth_model.php

protected $db;
.
.
.
$this->db = $this->load->database($this->config->item('database_group_name', 'ion_auth'), TRUE, TRUE);
```
If `database_group_name` is empty or `default` shouldn’t it recycle the same connection?

relevant commits https://github.com/benedmunds/CodeIgniter-Ion-Auth/pull/1194/commits/2f6c6eb9a508e99af5ba768488043bd30a076583 and merge https://github.com/benedmunds/CodeIgniter-Ion-Auth/pull/1194

Indigo744:
As discussed, Fix #1258 for branch 2

### DIFF
--- a/models/Ion_auth_model.php
+++ b/models/Ion_auth_model.php
@@ -183,18 +183,18 @@ class Ion_auth_model extends CI_Model
 		$this->lang->load('ion_auth');
 
 		// initialize the database
-        $group_name = $this->config->item('database_group_name', 'ion_auth');
-        if (empty($group_name)) 
-        {
-            // By default, use CI's db that should be already loaded
-            $CI =& get_instance();
-            $this->db = $CI->db;
-        }
-        else
-        {
-            // For specific group name, open a new specific connection
-            $this->db = $this->load->database($group_name, TRUE, TRUE);
-        }
+		$group_name = $this->config->item('database_group_name', 'ion_auth');
+		if (empty($group_name)) 
+		{
+			// By default, use CI's db that should be already loaded
+			$CI =& get_instance();
+			$this->db = $CI->db;
+		}
+		else
+		{
+			// For specific group name, open a new specific connection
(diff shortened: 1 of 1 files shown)


models/Ion_auth_model.php
```diff
@@ -183,7 +183,18 @@ public function __construct()
 		$this->lang->load('ion_auth');
 
 		// initialize the database
-		$this->db = $this->load->database($this->config->item('database_group_name', 'ion_auth'), TRUE, TRUE);
+		$group_name = $this->config->item('database_group_name', 'ion_auth');
+		if (empty($group_name)) 
+		{
+			// By default, use CI's db that should be already loaded
+			$CI =& get_instance();
```
benedmunds, line 190: Isn’t the CI super global already available as this here?

Indigo744: Yeah, but since we shadowed the $db member with our own, we need to access it using another mean if we want to set it...

benedmunds: I’m thinking maybe don’t override it at all with an empty group name. Or am I missing something?

Indigo744: I just verified, and as I suspect, not overriding when group name is empty leads to `Call to a member function select() on null`

Indigo744: This is because the `__get()` magic method from CI_Model will not be called if a `$db` member is defined in a child class (line 176).

benedmunds: Ah gotcha, ok merging this then.  Thanks!

Indigo744: No problem. It was good to check.

PR #1276, opened 2018-10-13
Fix TSQL syntax for UNIQUE constraint that can be NULL

### LINKED ISSUE #1255: Can't make a NULL column unique during v2 -> v3 SQL Server migration
**I get this error (excluded other 2):**

`Error in query (3621): The CREATE UNIQUE INDEX statement terminated because a duplicate key was found for the object name 'dbo.users' and the index name 'uc_activation_selector'. The duplicate key value is (<NULL>).
Could not create constraint or index. See previous errors.
The statement has been terminated.`

**from these last 3 commands:**

```
ALTER TABLE users
  ADD CONSTRAINT uc_activation_selector
    UNIQUE (activation_selector);

ALTER TABLE users
  ADD CONSTRAINT uc_remember_selector
    UNIQUE (remember_selector);

ALTER TABLE users
  ADD CONSTRAINT uc_forgotten_password_selector
    UNIQUE (forgotten_password_selector);
```

**This worked for me:**

> Pretty sure you can't do that, as it violates the purpose of uniques.
– https://stackoverflow.com/questions/191421/how-to-create-a-unique-index-on-a-null-column

`CREATE UNIQUE INDEX uc_activation_selector on users (activation_selector) where activation_selector is not null`

Indigo744:
Fix #1255

### DIFF
--- a/sql/ion_auth.mssql.sql
+++ b/sql/ion_auth.mssql.sql
@@ -20,13 +20,22 @@ CREATE TABLE users (
     phone varchar(20),
   PRIMARY KEY(id),
   CONSTRAINT uc_email UNIQUE (email),
-  CONSTRAINT uc_activation_selector UNIQUE (activation_selector),
-  CONSTRAINT uc_forgotten_password_selector UNIQUE (forgotten_password_selector),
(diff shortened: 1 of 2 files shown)


benedmunds:
Thanks!

Issue #1271, opened 2018-09-23
Activation codes are problematic to be used in a URL 


phalox:
An activation code, you would typically send via an email, and append to a URL.
But it seems that the algorithm used is generating characters that are not allowed:

e.g.
$2y$10$hSl57gPMjTfIPf0sv5ic3.WTC2nWpfbwRAqsFd8T/2xzjNWrBPgKi

$ signs are by default not allowed in a CI URL. Fixed by replacing with ":", and doing so again in the other direction.
/ (forward slash) is a valid output, and this causes the URL parameter to be broken up in 2 parts.

It also seems that the register() function is simply using this variable and adding it to the URL, which I believe will give issues in many cases?
From what I can see, this will be an issue in both hashing cases...

The easiest fix would be to use urlencode(), but it adds characters that make the URL look strange to the user. Or finding a different way for an activation key. In the end it's just used as an identifier that's stored entirely in the database. I don't see why you need special crypto for this. Just a good amount of length.

benedmunds:
Hmm, probably happens intermittently due to the random hash.  I think urlencoding is fine with me.  Anyone else have an opinion?

phalox:
I'm a little confused though. I've enabled the built-in email activation, but the code that's being passed in that URL is totally different from what's in the DB.

Here the codes:
![image](https://user-images.githubusercontent.com/3983738/46785339-be06ac00-cd31-11e8-9ba0-0e1c490bab17.png)

The email is generated with the default template and uses the id and activation_code of the ion_model. What am I missing here? The activation_code in the email seems to be more compatible with a well formatted URL.

PS: I'm creating a new user, then deactivating him so that they have to activate by email.

benedmunds:
Which branch are you using?  and I agree, it should use the users.activation_code

phalox:
I'm living somewhere on the V3 branch. I think on this commit:
https://github.com/benedmunds/CodeIgniter-Ion-Auth/commit/ec1822dd10fe20a30b4438e8a6f4a8ba1921582b

But if it'd use the activation_code we bump into the charachters issue. So
I'd really like to understand where this emailed code comes from.

The only place I see this used is:
Lib/Ion_auth line 299 and I don't see any conversion happen.

/me must be turning nuts

benedmunds:
@phalox check the forgotten_password_code and activation_selector for matches

phalox:
[...]

PR #1277, opened 2018-10-15
Explains activation code

Indigo744:
As discussed in #1271 

 - Added comment to explains better how `activation_code` works
 - Modify `activate()` function to extract the check activation code part, added as `check_user_code_activation()` function.
 - As a bonus, fix a bug in which event `post_activate_successful` would not be triggered if an activation code was provided

### DIFF
--- a/models/Ion_auth_model.php
+++ b/models/Ion_auth_model.php
@@ -46,7 +46,27 @@ class Ion_auth_model extends CI_Model
 
 	/**
 	 * activation code
-	 *
+	 * 
+	 * Set by deactivate() function
+	 * Also set on register() function, if email_activation 
+	 * option is activated
+	 * 
+	 * This is the value devs should give to the user 
+	 * (in an email, usually)
+	 * 
+	 * It contains the *user* version of the activation code
+	 * It's a value of the form "selector.validator" 
+	 * 
+	 * This is not the same activation_code as the one in DB.
+	 * The DB contains a *hashed* version of the validator
+	 * and a selector in another column.
+	 * 
+	 * THe selector is not private, and only used to lookup
+	 * the validator.
+	 * 
+	 * The validator is private, and to be only known by the user
+	 * So in case of DB leak, nothing could be actually used.
+	 * 
 	 * @var string
 	 */
 	public $activation_code;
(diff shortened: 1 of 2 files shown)


benedmunds:
This looks good to me.


@phalox can you take a look at this as well for any feedback you have

models/Ion_auth_model.php
```diff
@@ -360,52 +373,55 @@ public function rehash_password_if_needed($hash, $identity, $password)
...
-	public function activate($id, $code = FALSE)
+	public function check_user_code_activation($id, $code)
 	{
-		$this->trigger_events('pre_activate');
-
 		$token = $this->_retrieve_selector_validator_couple($code);
+	
+		$query = $this->db->select([$this->identity_column, 'activation_code'])
+							->where('activation_selector', $token->selector)
+							->where('id', $id)
```
phalox: I still don't fully get why the $id is used here, unless someone would be capable of changing the activation keys from one user to another, but that would mean they have access to the DB, so they could use the same key with any user.

Indigo744: Indeed, the selector would have been enough.

But it's again for the same reason... Backward compatibility! 😿

phalox: This function might be even more useful if it were to return the user object, but maybe that's a bridge too far?

If not, I could just use the $uid because it's been validated with the activation key in this function.

I'm not good with backwards compatibility, I'd break the entire world :-)

Indigo744: I understand.
You would have preferred the same type of function as `get_user_by_forgotten_password_code()`?

@benedmunds what do you think about modifying `check_user_code_activation()` to `get_user_by_activation_code()`?

benedmunds: I really think it's fine as is but I'm not against the change.  Or maybe add another method for get_user_by_activation_code that check_user_code_activation uses.

Indigo744: I didn't want to make it too much complicated, so I simply renamed it `get_user_by_activation_code()` and change its behavior to return the user on success. The id is no longer needed to get the user, but is still later checked in `activate()`.

The nice thing is that this new function is almost the same as `get_user_by_forgotten_password_code()`, which makes sense.

benedmunds: This looks good to me 👍

phalox:
That's looking great, and exactly what I need :-) Let's merge and I'll update my code!

benedmunds:
@Indigo744 is this good to merge? Anything else you want to change?

Indigo744:
@benedmunds No I'm good 😸

benedmunds:
Merged, thanks to both of you!

PR #1281, opened 2018-11-01
Identity check should url decode

jonlabelle:
... since the value stored in cookie is URL encoded.

### DIFF
--- a/models/Ion_auth_model.php
+++ b/models/Ion_auth_model.php
@@ -801,7 +801,7 @@ class Ion_auth_model extends CI_Model
 			return FALSE;
 		}
 
-		return $this->db->where($this->identity_column, $identity)
+		return $this->db->where($this->identity_column, urldecode($identity))
 						->limit(1)
 						->count_all_results($this->tables['users']) > 0;
 	}


benedmunds:
Can you post some example code.  


Couple things that stand out as possible issues:

identity_check is used in multiple places and they dont all pull the identity from a cookie.

I dont see anywhere that we're url encoding the identity when it's set in the cookie, so maybe that's happening somewhere else in your code or in CI?

jonlabelle:
Yes - the ci cookie helper will url encode the value. I will try to send over some examples later.

Also note, it seems to [here](https://github.com/benedmunds/CodeIgniter-Ion-Auth/blob/4a007e65cd6da0f14e68421ef91e20445e44a211/models/Ion_auth_model.php#L2118) (url decoding), but no where else.

benedmunds:
Weird, I dont see CI url encoding either, here's the call stack for that:
https://github.com/bcit-ci/CodeIgniter/blob/3.1-stable/system/helpers/cookie_helper.php#L70
->
https://github.com/bcit-ci/CodeIgniter/blob/3.1-stable/system/core/Input.php#L362
->
http://php.net/manual/en/function.setcookie.php


Looks like this urldecode was originally added with: 
https://github.com/benedmunds/CodeIgniter-Ion-Auth/commit/69728d192a0b1bba9363912330ebd9448d8a73c7

I wouldn't be against adding the urldecode to the getcookie() identity call to make sure all bases are covered, could be some CI/PHP version or setting that causes this.

Makes more sense to do it there, than in the method imho, since the method identity doesn't always come from a cookie.

jonlabelle:
So we may just want to consider taking out `urldecode` everywhere as it seems unnecessary, and [after reading](http://php.net/manual/en/function.setcookie.php#example-5630) the `setcookie` docs (which the CI Input calls [internally](https://github.com/bcit-ci/CodeIgniter/blob/fef6972a5cf19bcbec2ddc404cde6d5c4b432314/system/core/Input.php#L408))...

> *Note that the value portion of the cookie will automatically be urlencoded when you send the cookie, and when it is received, it is automatically decoded and assigned to a variable by the same name as the cookie name.*

benedmunds:
@avenirer you have any input here?  Dont want to revert it since you originally added it

avenirer:
Hello @jonlabelle and @benedmunds ,
Before doing this change, I started from a reported issue related to a remember me cookie problem. What I noticed was that, when trying to retrieve cookies that had an email address as value, the "@" sign was returned as an url encoded value. Looking now back at the CodeIgniter code (only a fugitive look), I see that, when xss_clean is enabled on the whole CodeIgniter app, the xss_clean function is applied to the cookie value just before it is returned (https://github.com/bcit-ci/CodeIgniter/blob/37d9b303163cd1256743d06750c1c742bd464823/system/core/Input.php#L171). Now, taking a look at xss_clean function it seems that it does some url_encode on the value (please, correct me if I am wrong; this was only a quick look at the code). That is why, at least in this particular case, a url_decode() seemed a good solution at the time. Take note, at that time, my only verification was an empirical one, so make sure that the elimination of the url_decode() in there doesn't affect the CodeIgniter app if the global xss clean is set to true (`$config['global_xss_filtering'] = TRUE;`)

Although the enabling of xss filtering globally is clearly discouraged in the documentation of CodeIgniter, it is still permitted. So, at least for the moment, even if it seems superfluous, maybe it would be a good idea to keep the url_decode() on that particular query...

benedmunds:
That makes sense @avenirer

Running url decide against a value that’s not encoded shouldn’t cause any issues anyway afaik, so I’m not against using it wherever we use getcookie to be safe

benedmunds:
Feel free to object if you think it would cause as issue

PR #1283, opened 2018-11-01
Corrected error on line 47 in lithuanian/ion_auth_lang.php

dgvirtual:


benedmunds:
Thanks for the catch there!

PR #1295, opened 2018-12-08
redirect with uri, not named/reverse-route

enix-app:
Based on the userguide:

> When passing a URI into the function, it is treated as a reverse-route request, not a relative/full URI, treating it the same as using redirect()->route():

```php
// Go to a named/reverse-routed URI
return redirect('named_route');
```


bvrignaud (approved):
You're right.
Thanks

Hello @benedmunds, can you merge this PR please ?
Thanks

benedmunds:
Merged

enix-app:
Thank you.

PR #1301, opened 2018-12-17
 Remove extra backslash 

tpw1314:
double backslash '\\' is not needed when running in console to seed. php spark will not find the correct seeder class.

### DIFF
--- a/INSTALLING.md
+++ b/INSTALLING.md
@@ -53,7 +53,7 @@ Don't forget to set Config\Migrations:enabled to true.
 
 You can also use the seeds file to insert default datas:
 ```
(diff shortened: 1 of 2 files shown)


bvrignaud:
As I said previously (https://github.com/bvrignaud/CodeIgniter-Ion-Auth/pull/1#issuecomment-447848423), that's not the case on my computer (linux). Can you add the 2 different ways please.

tpw1314:
> 
> 
> As I said previously ([bvrignaud#1 (comment)](https://github.com/bvrignaud/CodeIgniter-Ion-Auth/pull/1#issuecomment-447848423)), that's not the case on my computer (linux). Can you add the 2 different ways please.

Alright.

PR #1302, opened 2018-12-17
Spark seed - options for different OS

tpw1314:
Windows user may encounter 'Class IonAuthSeeder not found' error when trying to execute spark seed command in console. My conclusion is that Linux OS use double backslash `\\` when traversing whereas Windows use only single backslash.

example :
`php spark db:seed IonAuth\Database\Seeds\IonAuthSeeder`
vs
`php spark db:seed IonAuth\\Database\\Seeds\\IonAuthSeeder`

@bvrignaud as Linux user.

### DIFF
--- a/INSTALLING.md
+++ b/INSTALLING.md
@@ -53,7 +53,7 @@ Don't forget to set Config\Migrations:enabled to true.
 
 You can also use the seeds file to insert default datas:
 ```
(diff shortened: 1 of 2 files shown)


bvrignaud:
It's ok for me.
Thanks.

PR #1316, opened 2019-02-09
fix update group

candrasaputra:


benedmunds:
Hey, the previous code was correct but the docs were wrong until recently.  The docs have been updated here:
https://github.com/benedmunds/CodeIgniter-Ion-Auth/blob/3/userguide/index.html#L1243

PR #1324, opened 2019-02-28
Remove reference to ion_auth library inside ion_auth library for extending purposes

Stanzi1791:
I need to extend Ion Auth, because I need to add some additional functionality (two-factor authentication and an additional database unique constraint check) to it for a project I'm working on, and I don't want to modify the original code.

So, I created my own library class `Authentication extends Ion_auth` and autoloaded (in `autoload.php`):
- the `third_party` folder I put `Ion_auth` in (under packages)
- my extension library class, `authentication` and **not** the `ion_auth` library (under libraries)

This works fine for most of Ion_auth's functions, except for `deactivate`, because that function requires that the original `ion_auth` library itself is loaded and not an extension of it, see `Ion_auth_model.php` line 466: `$this->ion_auth->logged_in()`.

If I autoload only my extension library, without autoloading the `ion_auth` library, this gives me an `Undefined property Auth::$ion_auth` error for the `deactivate` function. Autoloading both libraries fixes that, but loads two nearly identical objects in memory (the original and the extension), which in my opinion isn't ideal.

To fix this, I've moved the `logged_in()` check in the `deactivate` function from the `Ion_auth_model.php` file to the `Ion_auth.php` file, removing the dependency on the original library class by replacing `$this->ion_auth->logged_in()`, with `$this->logged_in()`.

### DIFF
--- a/libraries/Ion_auth.php
+++ b/libraries/Ion_auth.php
@@ -265,7 +265,7 @@ class Ion_auth
 			}
 
 			// deactivate so the user must follow the activation flow
-			$deactivate = $this->ion_auth_model->deactivate($id);
+			$deactivate = $this->deactivate($id);
 
 			// the deactivate method call adds a message, here we need to clear that
 			$this->ion_auth_model->clear_messages();
--- a/libraries/Ion_auth.php
+++ b/libraries/Ion_auth.php
@@ -431,4 +431,22 @@ class Ion_auth
 		}
 	}
 
+	public function deactivate($id = NULL)
+	{
+		$this->trigger_events('deactivate');
+
+		if (!isset($id))
+		{
+			$this->set_error('deactivate_unsuccessful');
+			return FALSE;
+		}
+		else if ($this->logged_in() && $this->user()->row()->id == $id)
+		{
+			$this->set_error('deactivate_current_user_unsuccessful');
+			return FALSE;
+		}
+
+		return $this->ion_auth_model->deactivate($id);
+	}
+
 }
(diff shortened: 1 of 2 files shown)


benedmunds:
@Stanzi1791 good change here, thanks!  Should also change this reference to $this->deactivate() right?

https://github.com/benedmunds/CodeIgniter-Ion-Auth/blob/d36c4a9cb11ed425c8914133d0cdcef52b27c0e6/libraries/Ion_auth.php#L268

Stanzi1791:
I don't think that's neccessary, since you can be sure that the `ion_auth_model` is always loaded, since it is loaded by `Ion_auth` in it's constructor.

If you want to change that line, you'd also have to remove the loading of the `ion_auth_model` from the `Ion_auth` constructor. Which means you'd manually have to autoload both `ion_auth` and the `ion_auth_model` in order to use the library. Which doesn't seem right, since the model is not optional because it is always needed in order for `Ion_auth` to work, so I feel that `Ion_auth` should just load it.

Plus I can always create another model and load that in my `Authentication extends Ion_auth` class, and then use both the `ion_auth_model` and my own for database calls, so this is fine. :)

benedmunds:
@Stanzi1791 double check the context of the code I linked to.  I'm talking about when the register method calls deactivate.  We need it to call this->deactivate so it still triggers the events on register.

Stanzi1791:
Ah, I see, you are right, that needs to be updated too. I'll update the code. :)

benedmunds:
Awesome, thanks!

PR #1335, opened 2019-03-22
Update 20181211100537_install_ion_auth.php

robertkendrick:
Collate sequences on MySQL appear to be different to SQLite.
Edited to use the default collate sequences in both types of database system.


benedmunds:
Thanks!

PR #1338, opened 2019-04-03
Minor fixes on Auth.php

nichogx:
Fixed two issues:

1 - Phone and company were not required when creating a user, but were required when editing. It is now not required anymore on editing.
2 - Group description was not updating when updating a group. This pull request fixes this issue.


benedmunds:
Thanks!

PR #1339, opened 2019-04-03
Do not query database with null identity in recheck_session

najdanovicivan:
If user is not logged in when logged_in() method is called the following query is executed

```sql
SELECT `id`
FROM `users`
WHERE `email` IS NULL
AND `active` = '1'
ORDER BY `id` DESC
LIMIT 1 
```

This PR fixes the issue


benedmunds (commented):
Thanks for this!

PR #1343, opened 2019-04-06
Fix a portuguese translation

nichogx:
fix portuguese reset_password_validation_new_password_label

(was "Old Password" instead of "New Password")


benedmunds:
Thanks!

PR #1363, opened 2019-07-18
Change time() to now() (CI3 date helper)

EzraZebra:


benedmunds:
Anyone have any concerns about BC with this?

Only thing that comes to mind is that if CI time_reference is a different timezone than the PHP timezone some users might have session invalidation too soon or sessions lasting longer than they should.

EzraZebra:
Just posted this comment in the issue:
On second thought, time() is timezone independent (always UTC), whereas now() is not.
Changing it may cause unexpected results if the user isn't aware that now() is being used?

benedmunds:
Ah yea good point.  As long as we're consistent using time() shouldn't cause any issues right?  Did you hit an issue that caused you to look into this?

EzraZebra:
Sorry for the late reply.  
No, I was just looking through the code for a more or less unrelated reason and noticed that the date helper was being used, but not its now() function. I probably should have thought it through a bit more before submitting an issue report, sorry for any inconvenience.

benedmunds:
No worries, appreciate the contribution even if we aren't merging it.  Will close this, take care!

PR #1366, opened 2019-07-23
fixed 'get_user_by_activation_code' function: check token

wadealer:
We need to check token because of it can be `FALSE`

### DIFF
--- a/models/Ion_auth_model.php
+++ b/models/Ion_auth_model.php
@@ -385,16 +385,19 @@ class Ion_auth_model extends CI_Model
 	{
 		// Retrieve the token object from the code
 		$token = $this->_retrieve_selector_validator_couple($user_code);
-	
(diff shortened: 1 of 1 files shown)


benedmunds:
There's already a false check in the activate() method:
https://github.com/benedmunds/CodeIgniter-Ion-Auth/blob/3/models/Ion_auth_model.php#L418

Are you using this method separate from activate() or did you find that the false check in activate() isn't working properly?

wadealer:
Yes, I'm using this method separately. It's public so it should be ready for such usage.

wadealer:
However the check at `activate` method seems to be invalid. It's not checking `code` has two parts and can be splitted.

benedmunds:
Ok cool, makes sense.  Merging, thank you!

PR #1370, opened 2019-07-30
one more place where  should be checked

wadealer:
Here token can also be `FALSE` in case when `$user_code` has no '.'.


benedmunds:
Thanks!

PR #1384, opened 2019-11-12
Add support with CI4 database and save logging after register

MashinaMashina:


Models/IonAuthModel.php
```diff
@@ -558,20 +558,21 @@ public function changePassword(string $identity, string $old, string $new): bool
...
+						  ->select('id, password')
 						  ->where($this->identityColumn, $identity)
 						  ->limit(1)
 						  ->orderBy('id', 'desc')
-						  ->get($this->tables['users']);
-
-		if ($query->numRows() !== 1)
+						  ->get();
+		
+		if ($query->getFieldCount() !== 1)
```
MashinaMashina, line 568: Needs replace $query->getFieldCount() to $query->ResultID->num_rows in all lines

PR #1385, opened 2019-11-13
unify varying German translation of password

lordofbikes:
When I want to fix the varying German translation of password, I recognized the broken github link in all language files and fixed this too.


benedmunds:
Thank you!

PR #1386, opened 2019-11-15
remove reference to PHP >= 7.1 (CI4 do it for us)

bvrignaud:
It look like there was a typo error in the composer.json file.
Maybe It will be better to install it with composer.


benedmunds:
Thanks!

PR #1396, opened 2019-12-24
Update Auth.php

Mimisss:
Bugs descriped in issue https://github.com/benedmunds/CodeIgniter-Ion-Auth/issues/1397

**Proposed bug fixes**

**create_group**:
When `$new_group_id` is `FALSE` the error message(s) should be set and the create group form should be (re)displayed. Currently, a blank page is displayed if an error occurs since the `else` statements are never reached.

**edit_group**:
The user should be redirected to `auth`  only if the group update is successful otherwise any error message(s) should be displayed in the `edit_group` view.


benedmunds:
Awesome, thank you!  Really appreciate all the detail.

PR #1404, opened 2020-02-28
HOTFIX - #1403

### LINKED ISSUE #1403: Change password not working v4
# Hotifx 

## In the branch ci4, the method change_password is not working. This generates by the new structure of CI4

file Controller/Auth, line number 248

file Model/Ion_auth_model, method `change_password`

Daniel-Monroy:
It is resolved #1403

### DIFF
--- a/Controllers/Auth.php
+++ b/Controllers/Auth.php
@@ -245,7 +245,7 @@ class Auth extends \CodeIgniter\Controller
 		}
 		else
 		{
-			$identity = $this->session->userdata('identity');
+			$identity = $this->session->get('identity');
 
 			$change = $this->ionAuth->changePassword($identity, $this->request->getPost('old'), $this->request->getPost('new'));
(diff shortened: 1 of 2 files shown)


benedmunds:
Are these model changes needed as well even though they aren’t related to the session data?

Daniel-Monroy:
Yes, otherwise it produces an error

benedmunds:
Thanks!

Daniel-Monroy:
You're welcome, thanks for the great library Ben

benedmunds:
Thanks for the help!  Are you seeing any other issues? It looks like the other queries might need to change to use the builder structure as well.

Daniel-Monroy:
We are starting with CI4, if we had new problems as soon as possible, we
notified

benedmunds:
Thanks so much!

Issue #1414, opened 2020-03-14
Can't set username on registration


Max-02:
Hello
I just stated using ion auth so I got the newest version installed.
I am using the newest version of code igniter 3 and php 7.3.

In the Ion_auth_model.php register function is this on line 846:

$data = [
	$this->identity_column => $identity,
	'username' => $identity,
	'password' => $password,
	'email' => $email,
	'ip_address' => $ip_address,
	'created_on' => time(),
	'active' => ($manual_activation === FALSE ? 1 : 0)
];

Why do you set the username to the identity. So you cant set your username free. The username must be the identity. I want to set the email as the identity. In this case the username is the same as the email. 

I deleted this line of code: 'username' => $identity, and the bug was gone.

I may misunderstand the code, then I'm sorry. In any case, I would appreciate an answer.
Kind regards Max Klebe

benedmunds:
Pass in the email address as the username or identity in the register() call and it should work fine.  Identity is the name of the variable that’s passed in.

Max-02:
Yes, sorry if i get it wrong but i want an email as the key and a username that everyone can choose and the username should not be equal to the email.
I call the function like this: 

$password = $this->input->post("pass");
$email = $this->input->post("email");
$additional_data = array(
   'username' => $this->input->post("username"),
    'ID_Fachrichtung' => $this->input->post("specialization"),
 );
$this->ion_auth->register($email, $password, $email, $additional_data);

The config line:
$config['identity'] = 'email'; 

If I do this, the username is the same as the email.

benedmunds:
Ah ok. So for that pass the parameters like you are but pass the username under additional data on a subsequent update call. I think that will work, I’ve never seen someone want a username if it’s not the identify PK before actually...

Max-02:
Yes that shoud work.
This is your example of the register funciton on http://benedmunds.com/ion_auth/

$username = 'benedmunds';
$password = '12345678';
$email = 'ben.edmunds@gmail.com';
$additional_data = array(
      'first_name' => 'Ben',
       'last_name' => 'Edmunds',
 );
$group = array('1'); // Sets user to admin.
$this->ion_auth->register($username, $password, $email, $additional_data, $group);

But on stadard configuration and if you use the parameters right, the call of the register function should look like that:

$additional_data = array(
       'username' => $username,
      'first_name' => 'Ben',
       'last_name' => 'Edmunds',
 );
[...]

PR #1415, opened 2020-03-15
Delete a buggy line in the register function of the ion auth model

Max-02:
Let's say you want to register a user and the user should have a different email and username and the email is the identity.
This line of code, that I deleted, set the username equals to the identity and so equals to the email.
With deleting this line I prevented the described bug, but the user could use the username further as the identity.

### DIFF
--- a/models/Ion_auth_model.php
+++ b/models/Ion_auth_model.php
@@ -845,7 +845,6 @@ class Ion_auth_model extends CI_Model
 		// Users table.
 		$data = [
 			$this->identity_column => $identity,
-			'username' => $identity,
 			'password' => $password,
 			'email' => $email,
 			'ip_address' => $ip_address,


benedmunds:
Will it cause any issues to not have a username at all if the identity is email?

Max-02:
I tried a couple of functions and every one was working fine. Moreover, it wouldn't make any sense, because the functions I checked use the $identity_column and $identity to identify a user. The only function that of course needs a username is the check_username function, which works of course fine alter too.

benedmunds:
@Max-02 thanks!

Anyone else have input or concerns here?

bvrignaud:
https://github.com/benedmunds/CodeIgniter-Ion-Auth/blob/467f2bc6df56fc3c04b1b8bbec87482260a1c47e/config/ion_auth.php#L137
This config wasn't efficient ?

Max-02:
Its is, if you delete this line. 
If you watch the code, you can see that the username set to the identity even if you set your identity to email in the config.
```
$data = [
	$this->identity_column => $identity,
	'username' => $identity,
	'password' => $password,
	'email' => $email,
	'ip_address' => $ip_address,
	'created_on' => time(),
	'active' => ($manual_activation === FALSE ? 1 : 0)
];
```
The first entry of the data array use the identity config the others don't to this.

b3wii:
> 
> 
> Let's say you want to register a user and the user should have a **different email and username** and the email is the identity.
> This line of code, that I deleted, set the **username equals to the identity and so equals to the email**.
> With deleting this line I prevented the described bug, but the user could use the **username further as the identity**.

Hi @Max-02 I've not tested it thoroughly but dont think this is an error. Unfortunately I'm having difficulties to understand what you problem is (highlighted).

Max-02:
Maybe this helps you to understand: https://github.com/benedmunds/CodeIgniter-Ion-Auth/issues/1414

Max-02:
Hi @b3wii do you understand my issue now ?

b3wii:
@Max-02 Yes, I also don't understand what this line does. If username is identity than it is set correctly but if identity is something else and username field is present in form then value of identity field is saved to username column (which is definitely wrong). Username should be remove and also method username_check which is never used. It would also make this obsolete:

$lang['login_subheading'] = 'Please login with your email/username and password below.';

Max-02:
Yes, the username_check method isn't used in the ion auth code, but maybe some other devs use this method in their code. If you want to delete this method, you can delete the email_check method too. I agree with the $lang['login_subheading'] problem, but I think this is another issue.

Max-02:
Any other input or concerns?

b3wii:
@Max-02 Since the email property is required, we can keep email_check. Its not only login_subheading but also login_identity_label and forgot_password_identity_not_found which have the wrong text if you remove that username line.

Max-02:
@b3wii No, the text is wrong in any case.

- If you don't use the username, its still displayed as an option to enter.

- If you use the email as PK, and try to use the username at the login, the login will fail, because the login function use the identity field of the config. So it uses the email and the email is different to the username and it can't find the user.

So maybe you should add for example $lang['login_identity_label_username'] and $lang['login_identity_label_email'] and decide based on the config which one you display. The same on the other fields, that use the username/email.

By the Way sorry for my bad English and maybe difficult to understand sentences. I'm only 17 and from Germany, so English isn't my best skill.

b3wii:
@Max-02 Yes, the labels are wrong but this is not my repo..

benedmunds:
I'd rather not change labels and we can't remove methods since that would break BC.

Max-02:
Okay, deleting the methods or changing the labels is not necessary, but removing this line of code is.

Max-02:
Btw and what does BC means?

benedmunds:
BC = Backwards Compatibility

benedmunds:
Merging this, will watch for any issues it surfaces.  Thanks @Max-02 !

PR #1417, opened 2020-03-18
fix german phone_label translations

b3wii:
This problem also exists for other languages: https://i.imgur.com/V4ZpAxK.png


benedmunds:
Awesome, thank you!

Issue #1420, opened 2020-03-26
If flashdata is set to identity than the user is shown as logged in


b3wii:
@benedmunds What do you think of this code?

```
if($this->session->flashdata('identity') === $this->session->userdata('identity')){
	throw new Exception("Don't set flashdata identity!");
}
```

Because if flashdata is set to identity than user is falsely detected as logged in:

//reset flashdata & userdata

```
$this->session->set_userdata('identity', $somevalue);

var_dump($this->session->flashdata('identity')); => null
var_dump($this->session->userdata('identity')); => $somevalue
```

//reset flashdata & userdata

```
$this->session->set_flashdata('identity', $somevalue);

var_dump($this->session->flashdata('identity')); => $somevalue
var_dump($this->session->userdata('identity')); => $somevalue

$this->ion_auth->logged_in() => true
```

benedmunds:
How would this get set?  Are you worried that a developer would set that in the application code?

b3wii:
Yes, if you set flashdata before redirect and check after redirect for example in contructor if user is logged in. I found this because i worked on a project that didnt use the newest version where Ion_auth.php logged_in() returns this:

`return (bool) $this->ci->session->userdata($identity)`

But it could still happen in the current version if set_flashdata('identity') is used.

benedmunds:
Ah that makes sense. I’m cool with this check, send a PR whenever you get a chance please.

PR #1424, opened 2020-04-07
replace userdata('identity') check in recheck_session()

b3wii:
check for userdata('ion_auth_session_hash') and compare it to config value in order to determine if user is logged in


benedmunds:
Can you give some context on the reason for this change, who it affects, and what, if any, attack vectors it prevents?  Also, I'm interested in if this adds anything for applications with encrypted sessions.

b3wii:
> Can you give some context on the reason for this change, who it affects, and what, if any, attack vectors it prevents? Also, I'm interested in if this adds anything for applications with encrypted sessions.

The context is this: https://github.com/benedmunds/CodeIgniter-Ion-Auth/issues/1420

If $this->session->set_flashdata('identity') is set, user is not falsely logged in.

benedmunds:
Ah perfect, thank you!

PR #1428, opened 2020-04-17
corrections on translations

tgulum:
corrections on translations and add $lang['forgot_password_identity_not_found']


benedmunds:
Thank you!

PR #1431, opened 2020-04-19
Fixed email activation

Max-02:
In the Auth Controller I only removed a negation operator (the !).
Now it checks, if the $code string is empty to transfer the right parameter to the activate function.

In the Model I just changed 2 little things.
First I removed the bool datatype in the function head, that's let the function only accepts boolean.
I did that, because the function needs the code to work with and not just a boolean, whether the code is given or not.
Secondly I changed the comparative operator in a comparison from === to ==.
I don't know, if that's only a local problem, but on my system the $user->id variable is a string.
Because the $id variable is an int, the === operator return false.
If you changed this Operator, the comparison should work fine.

### DIFF
--- a/Models/IonAuthModel.php
+++ b/Models/IonAuthModel.php
@@ -374,13 +374,13 @@ class IonAuthModel
 	 * Validates and removes activation code.
 	 *
 	 * @param integer|string $id   The user identifier
-	 * @param boolean        $code The *user* activation code
+	 * @param string         $code The *user* activation code
 	 *                             if omitted, simply activate the user without check
 	 *
 	 * @return boolean
 	 * @author Mathew
 	 */
-	public function activate($id, bool $code=false): bool
+	public function activate($id, $code=false): bool
 	{
 		$this->triggerEvents('pre_activate');
 
@@ -390,7 +390,7 @@ class IonAuthModel
 		}
 		// Activate if no code is given
 		// Or if a user was found with this code, and that it matches the id
-		if ($code === false || ($user && $user->id === $id))
+		if ($code === false || ($user && $user->id == $id))
 		{
 			$data = [
 				'activation_selector' => null,
--- a/Models/IonAuthModel.php
+++ b/Models/IonAuthModel.php
@@ -431,7 +431,7 @@ class IonAuthModel
 			$this->setError('IonAuth.deactivate_unsuccessful');
 			return false;
 		}
-		else if ((new \IonAuth\Libraries\IonAuth())->loggedIn() && $this->user()->row()->id === $id)
+		else if ((new \IonAuth\Libraries\IonAuth())->loggedIn() && $this->user()->row()->id == $id)
 		{
 			$this->setError('IonAuth.deactivate_current_user_unsuccessful');
 			return false;
@@ -907,7 +907,7 @@ class IonAuthModel
 		{
 			if ($this->verifyPassword($password, $user->password, $identity))
 			{
-				if ($user->active === 0)
+				if ($user->active == 0)
 				{
 					$this->triggerEvents('post_login_unsuccessful');
 					$this->setError('IonAuth.login_unsuccessful_not_active');
--- a/Models/IonAuthModel.php
+++ b/Models/IonAuthModel.php
@@ -380,17 +380,17 @@ class IonAuthModel
 	 * @return boolean
 	 * @author Mathew
 	 */
-	public function activate($id, $code=false): bool
+	public function activate($id, ?string $code=null): bool
 	{
 		$this->triggerEvents('pre_activate');
 
-		if ($code !== false)
+		if ($code)
 		{
 			$user = $this->getUserByActivationCode($code);
 		}
 		// Activate if no code is given
 		// Or if a user was found with this code, and that it matches the id
-		if ($code === false || ($user && $user->id == $id))
+		if (is_null($code) || ($user && $user->id == $id))
 		{
 			$data = [
 				'activation_selector' => null,
--- a/Models/IonAuthModel.php
+++ b/Models/IonAuthModel.php
@@ -384,7 +384,7 @@ class IonAuthModel
 	{
 		$this->triggerEvents('pre_activate');
 
-		if ($code)
+		if (!is_null($code))
 		{
 			$user = $this->getUserByActivationCode($code);
 		}
--- a/Models/IonAuthModel.php
+++ b/Models/IonAuthModel.php
@@ -380,17 +380,17 @@ class IonAuthModel
 	 * @return boolean
 	 * @author Mathew
 	 */
-	public function activate($id, ?string $code=null): bool
+	public function activate($id, string $code=''): bool
 	{
 		$this->triggerEvents('pre_activate');
 
-		if (!is_null($code))
+		if ($code)
 		{
 			$user = $this->getUserByActivationCode($code);
 		}
 		// Activate if no code is given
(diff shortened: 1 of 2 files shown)


benedmunds:
Hey, can you add a description for what these changes are and why they're needed please.

Max-02:
BTW, if that isn't just a local problem with my database you should consider exchanging all the === with ==, because some other function use this too. For example the login function use it, to prove whether a user is active. I can login on an inactive account too on my system, if I use the === operator.

Models/IonAuthModel.php
```diff
@@ -374,13 +374,13 @@ public function getUserByActivationCode(string $userCode)
 	 * Validates and removes activation code.
 	 *
 	 * @param integer|string $id   The user identifier
-	 * @param boolean        $code The *user* activation code
+	 * @param string         $code The *user* activation code
```
bvrignaud, line 377: I Think `@param boolean|string $code ...` will be better

Max-02: I don't think so, because the function should only accept a string or nothing and nothing isn't a boolean.

bvrignaud: OK, so in this case, the function must be
`public function activate($id, ?string $code=null): bool`

And the follwing code is false : `$code === false`

Max-02: Was the ? a mistake?

bvrignaud: No : [https://www.php.net/manual/en/migration71.new-features.php](https://www.php.net/manual/en/migration71.new-features.php)

Max-02: Thanks for the information.
I committed a new version that contains your suggestion.

Models/IonAuthModel.php
```diff
@@ -374,23 +374,23 @@ public function getUserByActivationCode(string $userCode)
...
 	 * @return boolean
 	 * @author Mathew
 	 */
-	public function activate($id, bool $code=false): bool
+	public function activate($id, ?string $code=null): bool
 	{
 		$this->triggerEvents('pre_activate');
 
-		if ($code !== false)
+		if ($code)
```
benedmunds, line 387:
```suggestion
		if (!is_null($code))
```

Models/IonAuthModel.php
```diff
@@ -374,23 +374,23 @@ public function getUserByActivationCode(string $userCode)
...
 		{
 			$user = $this->getUserByActivationCode($code);
 		}
 		// Activate if no code is given
 		// Or if a user was found with this code, and that it matches the id
-		if ($code === false || ($user && $user->id === $id))
+		if (is_null($code) || ($user && $user->id == $id))
```
benedmunds: Are the id types different? Could we instead require an int as a typehint and then typecast the db?

Max-02: The $this->ionAuth->user()->row()->id returns a string.
So you could set a string as typehint for the $id, but I don't know what the advantage of it is against the ==.
And I don't know if this datatype is only a local problem with my windows or php version, because in the database the datatype of the id is an int.

Does anyone knows more about that?

benedmunds: Makes sense that it might change due to different DBs.  ID's could be strings too though if GUIDs or hashses are used.  So this seems fine to me.

benedmunds: @bvrignaud what are your thoughts? I'd prefer strict comparisons and typehints, maybe we should typecast to strings?

Max-02: Have strict comparisons any security benefits or why do you prefer it ?
BTW the deactivate function has the same problem with the datatypes.
And the login function use this code to prove if a user is active:
if ($user->active === 0) So inactive user can login too.
And maybe other functions have this problem too.

benedmunds: Prefer it so the types are explicit and clear.  I see your point though that this is probably an issue in quite a few places

benedmunds: What db engine are you running? MySQL/Postgres/etc

Max-02: MySQL with XAMPP and php version 7.4.4

benedmunds: @bvrignaud have you seen this or have any thoughts here?  I lean towards keep the explicit checks where possible and manually typecasting as needed so we can keep more explicit types throughout

bvrignaud: I agree.
I will do some small tests to answer you.

Models/IonAuthModel.php
```diff
@@ -374,23 +374,23 @@ public function getUserByActivationCode(string $userCode)
...
 	 * @param integer|string $id   The user identifier
-	 * @param boolean        $code The *user* activation code
+	 * @param string         $code The *user* activation code
 	 *                             if omitted, simply activate the user without check
 	 *
 	 * @return boolean
 	 * @author Mathew
 	 */
-	public function activate($id, bool $code=false): bool
+	public function activate($id, ?string $code=null): bool
```
bvrignaud:
```php
public function activate($id, string $code=''): bool
```

Models/IonAuthModel.php
```diff
@@ -374,23 +374,23 @@ public function getUserByActivationCode(string $userCode)
...
 	{
 		$this->triggerEvents('pre_activate');
 
-		if ($code !== false)
```
bvrignaud, line 387:
```php
if ($code)
```

Models/IonAuthModel.php
```diff
@@ -374,23 +374,23 @@ public function getUserByActivationCode(string $userCode)
...
+		if (!is_null($code))
 		{
 			$user = $this->getUserByActivationCode($code);
 		}
 		// Activate if no code is given
 		// Or if a user was found with this code, and that it matches the id
-		if ($code === false || ($user && $user->id === $id))
+		if (is_null($code) || ($user && $user->id == $id))
```
bvrignaud:
```php
if (! $code || ($user && $user->id == $id))
```

Models/IonAuthModel.php
bvrignaud: As I looked.
Actually the id is a int in database, and we havn't uuid.
So if dev change the users.id for an uuid, maybe we have to leave the id (in librarie/model) without "type hints" or force it to string.

Models/IonAuthModel.php
Max-02: That was your sugesstion, what is the benefit of this ?

Models/IonAuthModel.php
Max-02: `if (!is_null($code))` is the current code.

Models/IonAuthModel.php
Max-02: Yes, but event than has this condition `if ($user->active === 0)` form the login function the same problem.

Models/IonAuthModel.php
bvrignaud: not really, just to keep simple

Max-02: I agree, changed it.

Max-02:
I think the changing of the datatypes or to change the id to an uuid is a job for an additional PR.
So are there any other objections to the changes ?

benedmunds:
Agreed, thanks so much.  Merging!

datamweb:
@benedmunds , is this merged?
So why is there still a problem?
`https://github.com/benedmunds/CodeIgniter-Ion-Auth/tree/4`

benedmunds:
What problem are you having?

PR #1434, opened 2020-04-29
improve documentation

bvrignaud:
Improve routes documentation


benedmunds:
Is there any reason these are commented out in this example?

bvrignaud:
Yes.
I think we have to add/open route only for IonAuth functionalities that we will use.
So I just add these routes as examples.

benedmunds:
Makes sense.  Merging.  Thanks dude!

PR #1443, opened 2020-05-20
Only describe installing via Composer

benedmunds:
@bvrignaud I tend to agree with you, that all modern PHP apps should be using Composer.  So I'm proposing that we remove the other installing docs.  Thoughts?


bvrignaud:
Yes. Do it please.
But ... not sure, maybe it must be useful to leave some other install information for library development ?

benedmunds:
Good point, updated. Thanks!

Issue #1446, opened 2020-05-27
activation procedure


greald:
**Which branch are you using?**
4
**What commit hash are you on?**

**What CodeIgniter version are you using?**
v4

**What PHP version are you using?**
PHP 7.0

**Post your Ion Auth config below**
```
Config goes here
```

2020 May 27
I fixed several issues in the Ion Auth activation procedure.

\Libraries\IonAuth.php, function isAdmin()
fails if not logged in. Therefore
	original:
		`return  $this->ionAuthModel->inGroup($adminGroup, $id);`
	overruled by:
		`return ($this->loggedIn()) ? $this->ionAuthModel->inGroup($adminGroup, $id) : FALSE;`
		
\Controllers\Auth.php, function activate()
	original:

```
if (! $code)
		{
			$activation = $this->ionAuth->activate($id, $code);
		}
		else if ($this->ionAuth->isAdmin())
		{
			$activation = $this->ionAuth->activate($id);
		}

```
makes no sense.
	overturned into:
```
		if ($this->ionAuth->isAdmin())
		{
			$activation = $this->ionAuth->activate($id);
		}
		else
		{
			$activation = $this->ionAuth->activate($id, $code);
		}
```
		
\Models\IonAuthModel.php, function` activate($id, bool $code=false): bool`
original declaration never allows string $code.
	replaced by
		`function activate($id, $code=false): bool`

	original 

		`if ($code === false || ($user && $user->id === $id))`
fails because of unnecessary type mismatch.
	replaced by
```
// allow for type mismatches
		if ($code == false  || ($user && $user->id == $id))
```

benedmunds:
This is great, could you send a PR with these?

greald:
Yes Ben,
I can do that. I'm still working on the matter.
But I need a few days, may be week, for other business now.

:-)
Greald

Op wo 27 mei 2020 17:21 schreef Ben Edmunds <notifications@github.com>:

benedmunds:
Awesome, thank you!

PR #1447, opened 2020-06-02
fix isAdmin()

greald:
allow isAdmin check if not logged in.
see issue #1446

### DIFF
--- a/Libraries/IonAuth.php
+++ b/Libraries/IonAuth.php
@@ -377,7 +377,7 @@ class IonAuth
 
 		$adminGroup = $this->config->adminGroup;
 
-		return $this->ionAuthModel->inGroup($adminGroup, $id);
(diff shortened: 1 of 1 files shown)


Libraries/IonAuth.php
```diff
@@ -377,7 +377,7 @@ public function isAdmin(int $id=0): bool
 
 		$adminGroup = $this->config->adminGroup;
 
-		return $this->ionAuthModel->inGroup($adminGroup, $id);
+		return ($this->loggedIn()) ? $this->ionAuthModel->inGroup($adminGroup, $id) : FALSE;
```
benedmunds:
```suggestion
		return $this->loggedIn() && $this->ionAuthModel->inGroup($adminGroup, $id);
```

benedmunds: @greald does this change work for you?

benedmunds:
Thanks!

PR #1464, opened 2020-08-12
sometimes user creation may fail; in this case $id=0, we should check this

wadealer:


benedmunds:
Looks like insert_id wraps the underlying DBs insert return.  Which might not always give zeros but I think this should work fine across engines.

I'll leave this open for a few days to see if anyone has any objections to merging it?

benedmunds:
No objections so I’m merging this.  Thanks!

PR #1465, opened 2020-08-21
more security for "remember" cookie

wadealer:


benedmunds:
Thanks!

avenirer:
Wouldn't the secure flag mean that this will break the "Remember me" cookie of the websites that are not on "https://" ?

benedmunds:
Ah good point actually.  Since we’re using the cookie helper the secure flag should be set automatically based on HTTP/HTTPS already

benedmunds:
https://github.com/benedmunds/CodeIgniter-Ion-Auth/commit/149e30028133887d4ef1e41109cfc6a3f3ea6d39

wadealer:
@avenirer are you serious? No https in 2020? And site with authentication? It's might be a joke.

wadealer:
@benedmunds no it's not set, but should. Also this very important for SameSite="None" and Chrome.

benedmunds:
Ha you’d be surprised.  Especially for dev sites.  

Is it not setting for you dynamically?

wadealer:
Well, there is settings in config file, something like cookie_secure. I think we should respect this one.

benedmunds:
Yea, note https://github.com/bcit-ci/CodeIgniter/blob/899338a3ea4651533493ffd9d1ce7002e9ddf53e/system/core/Input.php#L391

PR #1474, opened 2020-09-16
Fix remember me

abitme:
fix remember me, when cookie destroyed after redirect, fix query in loginRememberedUser function, and fix cookie when logout


benedmunds:
Hey,

Did you find another solution to this?

abitme:
> Hey,
> 
> Did you find another solution to this?

Yeah, There's something that i have to add. I have sent my new pull request, please check it. hopefully it should works fine

benedmunds:
Ah ok thanks!

abitme:
I make new commit that keep $this->identityColumn in clearRememberCode function to avoid error when user doesn't use remember me

PR #1475, opened 2020-09-16
Fix remember me

abitme:


Models/IonAuthModel.php
```diff
@@ -504,6 +504,11 @@ public function clearRememberCode(string $identity): bool
 			'remember_code'     => null,
 		];
 
+		// get identity of logged in user
+		if ($identity == $this->identityColumn) {
+			$builder = $this->db->table($this->tables['users']);
+			$identity = $builder->select($identity)->where('id', $this->session->get('user_id'))->get()->getRowArray();
```
benedmunds: Why is this needed? We already have the identity and the user_id right?

abitme: when logout $identity is equal to $this->identityColumn so it produces like this email => email therefore it can't update table users to clear remember code.  I add this line of code to produce like this email => example@mail.com so it can update table users, is that make sense?

benedmunds: Could you do the update based on the user_id and remove one of the these queries?  Since the identify query is using the user_id anyway...

abitme: If you want to remove these queries then we should remove calling clearRememberCode in logout because it pointless can't update users. Remember code in users table will be cleared when user logout and then login without check remember me. how about that?

shoyshai: is the remember me working after doing this changes? Please update

abitme: You can see all changes in here : https://github.com/benedmunds/CodeIgniter-Ion-Auth/pull/1475/files

shoyshai: After making this changes still i remember me functionality is not working

benedmunds:
Any updates here?

benedmunds:
@shoyshai did you get this working?

abitme:
> Any updates here?

I have made some update, please check it.

shoyshai:
> 
> 
> @shoyshai did you get this working?

still remember me not working

abitme:
> > @shoyshai did you get this working?
> 
> still remember me not working

it's working fine on my pc. Can you explain the issue why it's not working?

benedmunds:
@shoyshai can you provide any more details?

abitme:
@benedmunds have you checked my pull request? is there's something wrong?

benedmunds:
I was hoping to hear back from @shoyshai before merging but this looks good to me. I'm going to merge and we can iterate later if any issues are surfaced.

Thanks for your time on this @abitme !

PR #1468, opened 2020-08-30
Modify forgot_password_identity_label

orbitasol:

PR #1478, opened 2020-10-05
Fix #1467, replace #1468

### LINKED ISSUE #1467: Language line "forgot_password_identity_label" in language/english/auth_lang.php is "forgot_password_username_identity_label" in other folders

lordofbikes:
This one goes back to febbc37 in Aug. 01. 2015.
As there are still 34 language files with `forgot_password_username_identity_label` and this is related with `forgot_password_email_identity_label`, I decided to rename `forgot_password_identity_label` back to `forgot_password_username_identity_label`.
Also identity can be username or email, so the old name makes sense for me.


benedmunds:
It might be safer to just do a forward change to update any un-updated language files.  Since this has been the label for ~5 years.

What do you think?

lordofbikes:
There are only 5 language files with the new label, but over 30 with the old label.
And because `identity` could be configured as `email` or `username` I found the old label more intuitive.
The label is always used in conjunction with `forgot_password_email_identity_label`, depending on configuration.

I've also checked branch 4, where the situation is still the same, 5 language files with `forgot_password_identity_label` and 33 files with `forgot_password_username_identity_label`.

But I think, that I understand your concerns now. When this issue was so long time undetected, I assume that localization is not used very much. But when English users build own templates and pages with the new label, the label rename will break them.
So I think now, that you're right, the opposite solution will be more safe.

When you agree, I can do it.

benedmunds:
Ok cool.  Yea that's exactly my concern, that or all of the people that built language files on the newest english file but just never contributed the translations back.

Agreed, thanks for this!

lordofbikes:
I reset my branch and made a forced push, so this should be fine now.
Will you merge this to branch `4` or should I create another pull request for it?

benedmunds:
Merged 🎉   If you can make another PR for v4 I'd appreciate it, thank you!

PR #1481, opened 2020-10-05
additional fix for Bengali language file

lordofbikes:
I additionally found this double index use in Bengali language file. After checking translated strings, the intention and the fix was clear.


benedmunds:
Thanks!

Issue #1354, opened 2019-05-29
Resending activation email


milad86:
Hello,

regarding activation email, how can we resend it ?
for example to allow a user to click a link to have the activation email sent again ??

benedmunds:
There's no built in way to resent it but you could write a controller method to do it, you can find the logic here:
https://github.com/benedmunds/CodeIgniter-Ion-Auth/blob/255c3dac483d57f9caf0946b8d9cfbbc57ffdfff/libraries/Ion_auth.php#L299

If you wanted to abstract this out so resends are possible I'd be happy to consider a PR for that.

strzala-dj:
I tried to make it as easy as possible, so I have implemented the resending of activation email feature by modifying `public function forgotten_password($identity)` in the Ion_auth.php library
https://github.com/benedmunds/CodeIgniter-Ion-Auth/blob/255c3dac483d57f9caf0946b8d9cfbbc57ffdfff/libraries/Ion_auth.php

Here it is:
```

	/**
	 * Forgotten password or resend activation email feature
	 *
	 * @param string $identity
	 *
	 * @return array|bool
	 * @author Mathew
	 */
	public function forgotten_password($identity)
	{
		// Retrieve activated user information
		$user = $this->where($this->ion_auth_model->identity_column, $identity)
					 ->where('active', 1)
					 ->users()->row();

		if ($user)
		{
			// Generate code
			$code = $this->ion_auth_model->forgotten_password($identity);

			if ($code)
			{
				$data = [
					'identity' => $identity,
					'forgotten_password_code' => $code
				];

				if (!$this->config->item('use_ci_email', 'ion_auth'))
				{
					$this->set_message('forgot_password_successful');
					return $data;
				}
				else
				{
					$message = $this->load->view($this->config->item('email_templates', 'ion_auth') . $this->config->item('email_forgot_password', 'ion_auth'), $data, TRUE);
					$this->email->clear();
					$this->email->from($this->config->item('admin_email', 'ion_auth'), $this->config->item('site_title', 'ion_auth'));
					$this->email->to($user->email);
					$this->email->subject($this->config->item('site_title', 'ion_auth') . ' - ' . $this->lang->line('email_forgotten_password_subject'));
					$this->email->message($message);

					if ($this->email->send())
					{
						$this->set_message('forgot_password_successful');
						return TRUE;
					}
				}
			}
		}



		// Retrieve not activated user information
[...]

PR #1506, opened 2021-03-29
Update IonAuthModel.php

DrAliRagab:
I think you should have function to resend activation email if the first one was not sent successfully.
Depending on that [code ](https://github.com/benedmunds/CodeIgniter-Ion-Auth/issues/1354)I think I made a working function.

The only problem I can't get identity to be sent to email activation, I tried but I failed.
I don't know how to get the identity field.
Actually I didn't tried that hard because I don't need it at activation email, I just want first and last name.


benedmunds:
Thanks for this PR, this seems like a great addition.  Here are a couple of changes to consider:

1. Move this method to the library, since it's where the other email logic lives and it's not data related
2. Remove deactivation from this method, it should only send the email (if the user is deactivated)
3. Rename to SendActivationEmail, and refactor the register method to call this method instead of sending it's own email directly

To resend emails I would expect the controller logic to be something like `$this->ionAuth->deactivate($user->id)->sendActivationEmail($user->identity)`.

Also, please combine the changes from #1507 into this PR.  I'll be closing that one.

DrAliRagab:
I think it is better to keep register and send email activation to separate functions.
And we should keep deactivate so we change the activation link every new request.

Libraries/IonAuth.php
```diff
@@ -292,6 +292,86 @@ public function register(string $identity, string $password, string $email, arra
...
 
+	/**
+	 * Send activation email.
+	 *
+	 * @param string $identity
+	 *
+	 * @return boolean|array return an array of activation details if CI e-mail validation is enabled
+	 * @author Ali Ragab
+	 */
+	public function SendActivationEmail(string $identity): bool|array
```
benedmunds:
```suggestion
	public function sendActivationEmail(string $identity): bool|array
```

Libraries/IonAuth.php
```diff
@@ -292,6 +292,86 @@ public function register(string $identity, string $password, string $email, arra
...
+
+		if ($user->active) {
+			$this->setError("IonAuth.already_activated_identity");
+			return FALSE;
+		}
+
+		if ($user) {
+
+			// deactivate so the user must follow the activation flow
+			$deactivate = $this->deactivate($user->id);
```
benedmunds: Is this (and the next few lines) needed? You're throwing an error above if the user is active.

benedmunds (changes requested):

Libraries/IonAuth.php
DrAliRagab: You are right.
Not logic to keep `if ($user) {` as we checked the identity against database and the user must be founded.

But my point regarding `$deactivate = $this->deactivate($user->id);` is to change the activation code every time we request activation code sending. I think it will be more secure to change activation code every request. Am I right?

May be we can regenerate the token and update database, like this code from \Models\IonAuthModel.php
```php
$token                = $this->generateSelectorValidatorCouple(20, 40);
		$this->activationCode = $token->userCode;

		$data = [
			'activation_selector' => $token->selector,
			'activation_code'     => $token->validatorHashed,
			'active'              => 0,
		];
```

But deactivate function do this job perfectly.

benedmunds: Good point.  You could call the model's deactivate method directly to separate those concerns.

benedmunds:
This is looking great. 

One last thing, can you refactor the register() method to use sendActivationEmail(), that way we're reducing the duplicated code there.

DrAliRagab:
While working I want to send email to user if his password changed successfully.
So I wrote `sendEmail()` function.
I think it will decrease the duplicate and may be used for further expansion of this Auth system.
My point is that `Register()` trigger defferent events than `sendActivationEmail()`, So better to keep sending email separate.

DrAliRagab:
While working I want to send email to user if his password changed successfully.
So I wrote `sendEmail()` function.
I think it will reduce duplicated code here. Also, may be useful for further expansion of this Auth system.

Libraries/IonAuth.php
```diff
@@ -270,16 +270,14 @@ public function register(string $identity, string $password, string $email, arra
...
-
-				$this->email->clear();
-				$this->email->setFrom($this->config->adminEmail, $this->config->siteTitle);
-				$this->email->setTo($email);
-				$this->email->setSubject($this->config->siteTitle . ' - ' . lang('IonAuth.emailActivation_subject'));
-				$this->email->setMessage($message);
-
-				if ($this->email->send() === true)
-				{
+				$email_sent = $this->sendEmail(
```
benedmunds:
```suggestion
				$emailSent = $this->sendEmail(
```

Libraries/IonAuth.php
```diff
@@ -292,6 +290,113 @@ public function register(string $identity, string $password, string $email, arra
...
+		];
+
+		if (!$this->config->useCiEmail) {
+			$this->ionAuthModel->triggerEvents(['activation_email_successful']);
+			$this->setMessage('IonAuth.activation_email_successful');
+			return $data;
+
+		} else {
+
+			$email_sent = $this->sendEmail(
```
benedmunds:
```suggestion
			$emailSent = $this->sendEmail(
```

Libraries/IonAuth.php
```diff
@@ -292,6 +290,113 @@ public function register(string $identity, string $password, string $email, arra
...
+				$user->email,
+				$this->config->siteTitle . ' - ' . lang('IonAuth.emailActivation_subject'),
+				$this->config->emailTemplates . $this->config->emailActivate,
+				$data
+			);
+
+			if ($email_sent) {
```
benedmunds:
```suggestion
			if ($emailSent) {
```

Libraries/IonAuth.php
```diff
@@ -270,16 +270,14 @@ public function register(string $identity, string $password, string $email, arra
...
-				if ($this->email->send() === true)
-				{
+				$email_sent = $this->sendEmail(
+					$email,
+					$this->config->siteTitle . ' - ' . lang('IonAuth.emailActivation_subject'),
+					$this->config->emailTemplates . $this->config->emailActivate,
+					$data
+				);
+
+				if ($email_sent) {
```
benedmunds:
```suggestion
				if ($emailSent) {
```

Libraries/IonAuth.php
```diff
@@ -292,6 +290,113 @@ public function register(string $identity, string $password, string $email, arra
...
+		$this->email->clear();
+		$this->email->clear();
+		$this->email->setFrom($this->config->adminEmail, $this->config->siteTitle);
+		$this->email->setTo($user_email);
+		$this->email->setSubject($subject);
+		$this->email->setMessage($message);
+
+		if ($this->email->send() === true) {
+			return TRUE;
+		}else{
```
benedmunds:
```suggestion
		} else {
```

Language/en/IonAuth.php
```diff
@@ -68,6 +68,9 @@
 	'emailActivate_heading'              => 'Activate account for %s',
 	'emailActivate_subheading'           => 'Please click this link to %s.',
 	'emailActivate_link'                 => 'Activate Your Account',
+	'empty_identity'          			 => 'Some thing wrong. Please, try again.',
```
benedmunds:
```suggestion
	'empty_identity'          			 => 'Something went wrong. Please try again.',
```

benedmunds (changes requested):
Just a couple small edits and this should be ready to merge, thanks so much!

Libraries/IonAuth.php
```diff
@@ -292,6 +290,113 @@ public function register(string $identity, string $password, string $email, arra
...
+	public function sendEmail(
+		string $user_email, 
+		string $subject, 
+		string $template, 
+		array $data): bool
+	{
+		$message = view($template, $data);
+
+		$this->email->clear();
+		$this->email->clear();
```
benedmunds:
```suggestion
```

benedmunds (changes requested):
Found one more issue.  After this I'll merge.

DrAliRagab:
Good review.

benedmunds:
Thanks!

PR #1519, opened 2021-05-19
Update IonAuthModel.php

datamweb:
Change Boolean to Bool because:
"Boolean" is interpreted as a class.
CI : 4.1.2
PHP Version : 8.0.3


benedmunds:
Thanks!

PR #1523, opened 2021-05-27
fix activation_code 

datamweb:
Hi,
set `public $emailActivation          = true;               	// Email Activation for registration`
in Config/IonAuth.php file.
Then create a new user (auth/create_user).
Existence of an error after creating a new user.
Shouldn't it " activationCode "?


benedmunds:
Thanks!

PR #1530, opened 2021-07-07
fixbugs ci4 filterData postgresql 

junglaCODE:
the problem is in getFieldNames function. Postgres disconnects and returns nothing and the insert does not take the extra data.

https://forum.codeigniter.com/thread-65675.html


junglaCODE:
hey ho lest go

Issue #1528, opened 2021-07-03
filterData


junglaCODE:
**Which branch are you using?**
2/3/*4 (branch used should match CodeIgniter version)

**What commit hash are you on?**
Attempt updating to the latest commit on your branch.

**What CodeIgniter version are you using?**
v2/v3/*v4

**What PHP version are you using?**
PHP 5.4/5.6/5.7/7.0/*7.4/etc

**Describe the bug**

FilterData no it does not work getFieldNames , I think the error is related to https://forum.codeigniter.com/thread-65675.html

thanks

junglaCODE:
the database is posgresql 12

benedmunds:
Can you confirm that you're on the latest commit of CI?

junglaCODE:
yes is the latest commit, i belive problem  it's related to https://forum.codeigniter.com/thread-65675.html.  I used postgresql 12 . I hope I can solve it and send the patch. Thanks. I am going to try mysql

junglaCODE:
Solve this problem with this modification in the function. I hope it is helpful for this library.  :)  #developersjunglaCode   From  Villagran Guanajuato

  /**
	 * Filter data
	 *
	 * @param string $table Table
	 * @param array  $data  Data
	 * @author jlgc/monolinux
	 * @return array
	 */
	protected function filterData(string $table, $data): array
	{
			$filteredData = [];
			if( $this->db->DBDriver === 'Postgre'){
				$structTable  = $this->db->query("SELECT column_name  FROM information_schema.columns     
				WHERE table_schema = 'security'    
				AND table_name   = 'users'")->getResult();	
				$columns = null;	
				foreach($structTable  as $key => $value ){
					$columns[] = $value->column_name;
				}
			}else{
				$columns = $this->db->getFieldNames($table);
			}
			if (is_array($data))
			{
				foreach ($columns as $column)
				{	
					if (array_key_exists($column, $data))
					{
						$filteredData[$column] = $data[$column];
					}
				}
			}
			return $filteredData;
	}
Solveted this problem

benedmunds:
Can you submit a new PR with just this change?  Looks like the other PR is against the master branch instead of the 4 branch so I'm going to close that one.

benedmunds:
Closing due to inactivity.  Please feel free to re-open when the PR is ready for another review.

PR #1532, opened 2021-07-08
fixbug relacionado con el tema de filterdata #1528

junglaCODE:
I think I struggle a lot. finally I think I stay. Greetings bened from Villagran Gto, Mexico


benedmunds:
Hey man, this is closer! But it has a lot of edits that aren't related to this change.  Looks like it's including whitespace and editor changes.

You might want to do another branch with only this change.

benedmunds:
Closing due to inactivity.  Please feel free to re-open when this is ready for another review.

PR #1541, opened 2021-10-22
Updated portuguese language files

mpmont:
- Changed the "usuario" term to "utilizador" a more common term in Portuguese. 
- Translated a few keys that were still missing
- Corrected a typo in the recover password link


benedmunds:
Thanks!

PR #1545, opened 2022-01-04
correction var type and function

sudadi:


benedmunds:
Thanks! These changes look great but can you fix the intention on the numRows code?

PR #1549, opened 2022-02-07
Update japanese translation

sunadarake:
Thanks for the great library!
This PR attempts to update Japanese Translation files.

- Translate a few keys into Japanese
- Fix indents in Japanese translation files


benedmunds:
Thanks!

Issue #1551, opened 2022-02-27
install with composer problem


nemanjagmitric72:
I have a problem installing IonAuth with composer. 
When i run comand "composer require benedmunds/CodeIgniter-Ion-Auth:4.x-dev" i have error:  "./composer.json has been updated
Installation failed, reverting ./composer.json and ./composer.lock to their original content.
[RuntimeException]
  require.benedmunds/CodeIgniter-Ion-Auth is invalid, it should not contain uppercase characters. Please use benedmunds/codeigniter-ion-auth instead.  "
Can you help me whit this problem?

benedmunds:
This has been fixed

PR #1552, opened 2022-03-12
fix for composer

avenirer:
Writing the branch with uppercase letters would end with a "[RuntimeException] require.benedmunds/CodeIgniter-Ion-Auth is invalid, it should not contain uppercase characters. Please use benedmunds/codeigniter-ion-auth instead." This should sovel the issue #1551


benedmunds:
Thanks!

PR #1553, opened 2022-03-12
Better explaining of autoloading IonAuth

avenirer:
I assume this would explain better where the line should be inserted


benedmunds:
Thanks!

PR #1564, opened 2022-05-28
Correct reference to the manual file

dgvirtual:
The external link still points to the version 3 of the manual, I think the link to correct version would be helpful.


benedmunds:
Thanks!

PR #1566, opened 2022-06-08
Update Auth.php

dgvirtual:
Remove trim function from the validation rules as "[CI4 validation never changes your data](https://codeigniter.com/user_guide/installation/upgrade_validations.html)".


benedmunds:
We should still trim these in the app code right?

dgvirtual:
I guess so, but trimming should be done before validation. If we validate untrimmed identity, for example, and the user happens to leave a space at the end, the identity will validate despite such identity already existing, right?

benedmunds:
Yep agreed

setioaji:
need trim

PR #1568, opened 2022-07-27
Refactor model

orionstar:


controllers/Auth.php
```diff
@@ -780,7 +772,7 @@ public function create_group()
 	public function edit_group($id)
 	{
 		// bail if no group id given
-		if (!$id || empty($id))
+		if (empty($id))
```
benedmunds, line 775: Should this be an isset instead?

benedmunds (changes requested):
Thanks for the work on this.

Overall these changes look good but there's a lot going on here for one PR.  Could you break this up into several more focused PRs?

orionstar:
I'll break it into multiple pr/branch

Issue #1570, opened 2022-09-23
Users can still log in using "Remember Me" cookies when the feature was disabled


chland:
**Which branch are you using?**
3

**What commit hash are you on?**
latest, downloaded today (e47b492 i guess)

**What CodeIgniter version are you using?**
v3.1.13 (latest release)

**What PHP version are you using?**
7.2.31

**Post your Ion Auth config below**
Pretty much the default one with only some table-names changed.

**Describe the bug**
The `logged_in()` function doesn't take the value of  `$config['remember_users']` into account when checking for a "remember_me" cookie.

If you enable the "remember me" feature and a user logs in, he gets logged in the next time he visits your site, even if you set  `$config['remember_users']`  to `false` in the meantime.

**To Reproduce**
1. set `$config['sess_expiration']` to `0` in `config.php`
2. set `$config['remember_users']` to `true` in `ion_auth.php`
3. log in to your site, with "Remember Me" enabled
4. close your browser
5. open your browser and open the site again -> you are logged in (which is correct at this point)
6. close your browser again
7. set `$config['remember_users']` to `false` in `ion_auth.php`
8. open your browser and your site -> you are logged in - which should NOT happen

**Expected behavior**
The result of step 7 should be that the next time you visit the site,, the "remember me" cookie is ignored and you're not logged in, even if a valid cookie exists.

I did a very quick test and it seem that this behaviour is caused by this bit of code in the logged_in() function:

```
		// auto-login the user if they are remembered
		if (!$recheck && get_cookie($this->config->item('remember_cookie_name', 'ion_auth')))
		{
			$recheck = $this->ion_auth_model->login_remembered_user();
		}

```

which only checks the name of the cookie but doesn't take into account if  `$config['remember_users']` is true or false. IMHO the code should look like this:

```
		// auto-login the user if they are remembered
		if (!$recheck && ($this->config->item('remember_users', 'ion_auth')) && get_cookie($this->config->item('remember_cookie_name', 'ion_auth')))
		{
			$recheck = $this->ion_auth_model->login_remembered_user();
		}
```

benedmunds:
Thanks for the detailed report!  This makes sense, can you submit a PR with this change?

PR #1571, opened 2022-09-28
"Remember Me" Improvement

chland:
Fixes an edge case (described in Ticket #1570) where people were able to login  if the "Remember Me" feature was disabled in the app but people still had old "Remember Me" cookies.


benedmunds:
Thanks!

PR #1577, opened 2023-03-13
don't get $user twice; some code cleanup

wadealer:
Removed some extra spaces.
`$user = $this->user($id)->row();` was called twice at `update` method.

### DIFF
--- a/models/Ion_auth_model.php
+++ b/models/Ion_auth_model.php
@@ -46,27 +46,27 @@ class Ion_auth_model extends CI_Model
 
 	/**
 	 * activation code
-	 * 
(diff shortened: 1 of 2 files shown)


models/Ion_auth_model.php
```diff
@@ -1810,7 +1810,7 @@ public function update($id, array $data)
 			{
 				if( ! empty($data['password']))
 				{
-					$user = $this->user($id)->row();
+//					$user = $this->user($id)->row();
```
benedmunds, line 1813: Did you intend to remove one of these and then comment out the other?

PR #1588, opened 2024-09-10
Add codeigniter framework version back to composer requirement

canihavesomecoffee:
In aba58c498d494f2d8e064e9fd85c4aba473c5e64 the reference to CodeIgniter was removed as it was in a dev version for a long while.

In the last 5 years however CI was released under the 4.x branch, so it seems apt to add it again.


benedmunds:
Thanks!

